Issue #38, opened 2025-05-15
Full Bash Script to run everything?


joshua-quek-sonarsource:
Hi @perry-sonar @SonarFox , I generated a full script to run this. 

Was wondering if it looks correct, and if it is, if it might be worth adding this to the Readme?

```bash
#!/bin/bash

# ==================================================================================
# SonarQube Migration Script
# ==================================================================================
# This script automates the process of migrating data from SonarQube Server to 
# SonarQube Cloud using the sonar-reports tool.
#
# Prerequisites:
# - Docker installed and running
# - Access to both source SonarQube Server and target SonarQube Cloud
# - Admin tokens for both instances
# ==================================================================================

# User-configurable variables
# ------------------------------------------------------------------------------------------
# Source SonarQube Server information
SOURCE_URL="https://your-sonarqube-server-url"  # URL of the source SonarQube server
SOURCE_TOKEN="your-source-token-here"           # Admin user token for the source SonarQube server

# Certificate information (if needed)
PEM_FILE_PATH=""                                # Path to client certificate pem file
KEY_FILE_PATH=""                                # Path to client certificate key file
CERT_PASSWORD=""                                # Password for client certificate

# Extract options
EXPORT_DIRECTORY="./sonarqube-export"           # Directory to store exported data
EXTRACT_TYPE=""                                 # Type of Extract to run (leave empty for default)
CONCURRENCY=4                                   # Maximum number of concurrent requests
TIMEOUT=60                                      # Request timeout in seconds
EXTRACT_ID=""                                   # ID of an extract to resume in case of failures

# Target SonarQube Cloud information
CLOUD_TOKEN="your-cloud-token-here"             # Admin token for SonarQube Cloud
ENTERPRISE_KEY="your-enterprise-key"            # Enterprise key in SonarQube Cloud
CLOUD_URL="https://sonarcloud.io"               # SonarQube Cloud URL
EDITION="enterprise"                            # SonarQube Cloud Edition

# Migration options
RUN_ID=""                                       # ID of a run to resume in case of failures
TARGET_TASK=""                                  # Name of a specific migration task (leave empty for all tasks)

# Docker image
DOCKER_IMAGE="ghcr.io/sonar-solutions/sonar-reports:latest"

# ------------------------------------------------------------------------------------------
# Script execution starts here
# ------------------------------------------------------------------------------------------

# Check if Docker is installed and running
if ! command -v docker &> /dev/null; then
[...]

Issue #30, opened 2025-01-28
Create a new command to just get the report


arturo-palau-sonarsource:
For Onboarding and Support purposes, having just a command to execute so that the report is generated would be great to have. That way the Onboarding and Support teams can provide this and it will be pretty direct for the customers to just run the command instead of having to run the extract and then report commands to generate the report. 

I would this something like:
sonar-reports report-only URL USER_TOKEN

Thanks!

PR #43, opened 2025-09-03
Bash script to create reports for quicker assessment

michail-melizanis-sonarsource:
[Draft PR for now]

Introducing into this PR:
1. #38: : Incorporated -and improved a bit- @joshua-quek-sonarsource  bash script (startup.sh)
    - #30: Added also the ability to have the bash script stop after getting the initial report
2. Included inside the report (migration.md) a column for Last Analysis Scan and order the Projects with it

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,0 +1,380 @@
+#!/bin/bash
+
+# ==================================================================================
+# SonarQube Migration Script
+# ==================================================================================
+# This script automates the process of migrating data from SonarQube Server to 
+# SonarQube Cloud using the sonar-reports tool.
+#
+# Prerequisites:
+# - Docker installed and running
+# - Access to both source SonarQube Server and target SonarQube Cloud
+# - Admin tokens for both instances
+#
+# ⚠️  SECURITY WARNING: 
+# - NEVER commit real tokens or URLs to version control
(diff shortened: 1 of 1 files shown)


michail-melizanis-sonarsource:
Alright, now everything should be more clear and I actually remember why I changed them!

-  The problem with Project Metrics' table in the report, which it wasn't get populated, it was because the filter in `src/report/common/projects.py` was too restrictive and it was only including projects with custom template rules or plugin rules (`filter_lambda=lambda x: x['template_rules'] > 0 or x['plugin_rules'] > 0`). I've changed this to just `filter_lambda=lambda x: True `. 
I hope this should be fine though, and those projects wasn't getting treated differently on the migration phase!
- In the `migration/generate.py` we were discarting data with `*_` and now instead we're capturing them. 

The rest of the commits should be just for the additional column of having the Last Analysis Scan! 

The rest of the commits on the previous PR didn't seem to add anything of value, and they were probably coming from my tests on trying to automate the migration 😬 sorry for this! 

Let me know if you think that those changes are fine, and we should be ready to go 🚀

perry-sonar (approved):
thanks for being the first to contribute!

PR #45, opened 2026-02-04
Updated README file to improve documentation

kiloecho-sonar:
I used claude to change the format of the README file to the more widely utilized markdown format.  
I added a rudimentary workflow diagram, fixed a few mistakes with the order of arguments/options from the original README, and improved the readability to reduce user error/confusion.  
This will be updated again as I improve the tool.


joshua-quek-sonarsource:
Hi @kiloecho-sonar , thanks for the swift reply. I believe that you are referring to version of the `README.md` from commit `aa8fe0938a269e87d92666dd6904a38decc7cf09` that was made on 4th Sept last year. 

Moving forward, I was not referring to that version of the `README.md` but was referring to the most recently `README.md` on commit `4a941afaa3c6b07df5221a62f706a34803080d74` that the merge conflict is happening on. 

As per what @okorach-sonar @jon-lipsky-sonarsource and myself have been doing in terms if clean gitops practices before a merge, would it be possible to do a rebase onto main branch first, before we take a look at this again?

Issue #109, opened 2026-05-18
`sonar-migration-tool` apparently stuck on `setProjectTags` task


okorach-sonar:
I am working from the main branch build locally as of Monday May 18th morning.
My migration gets stuck on `setProjectTags`
```
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=addMigrationGroupToTemplates
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=addMigrationUserToMigrationGroups
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=createPortfolios
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=getGateConditions
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=getProfileBackups
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=getProjectIds
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=setDefaultTemplates
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=setOrgGroupPermissions
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=setProfileGroupPermissions
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=setProfileParent
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=setProjectGates
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=setProjectGroupPermissions
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=setProjectProfiles
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=setProjectSettings
time=2026-05-18T14:16:19.715+02:00 level=INFO msg="running task" task=setProjectTags


<stuck since 15 minute on this step>
```

kiloecho-sonar:
The issue is that there are multiple subroutines running at the same time but a semaphore pool of only 25 available threads.  Each subroutine requires 2 threads so we run into a deadlock where all subroutines grab all of the available threads and then try to grab the second thread (which the pool is now exhausted).  Thanks to a log from @daniel-borek-sonarsource we were able to detect the root cause and claude is changing this from semaphore pool to errgroup.SetLimit() for concurrent processing.  This will be available in the kilo branch for testing.

[...]

PR #121, opened 2026-05-19
Issues task #109: sonar-migration-tool apparently stuck on setProjectTags

kiloecho-sonar:
Fix migration deadlock: replace semaphore with errgroup.SetLimit. 

Double semaphore acquisition in read task callbacks caused deadlock when goroutines held all 25 slots and each needed a second slot. Replace manual AcquireSem pattern with errgroup.SetLimit for bounded concurrency. Also adds parallel processing to 7 previously sequential tasks for enterprise scale.


kiloecho-sonar:
I will fix coverage and duplicates on the next commit.

kiloecho-sonar:
> I will fix coverage and duplicates on the next commit.

Moving this to the end so reviewer sees it

jon-lipsky-sonarsource (approved):
👍 from me.

PR #466, opened 2026-07-03
fix(ci): fail clearly on Vault miss in signing; document required Vault access

joshua-quek-sonarsource:
## Why

The first live run of the tag-triggered release pipeline (tag `v1.0.0`, Actions run [28650002595](https://github.com/SonarSource/sonar-migration-tool/actions/runs/28650002595)) failed. `Test` and all six `Build All Binaries` passed (cross-compiled + GPG-signed), but **`Sign and Notarize macOS Binaries` failed**, so `Publish Release` was skipped and **no release was created**.

Root cause is Vault permissions, not code: the GitHub-hosted macOS runner reaches internal Vault via Cloudflare WARP, and the repo's Vault role `github-SonarSource-sonar-migration-tool-protected` was **denied** read on `development/kv/data/cloudflare/warp-github-runner`:

```
##[error]  DENIED  development/kv/data/cloudflare/warp-github-runner
##[error]Vault secret access denied for 1 path(s).
```

Because the Vault step then returned no output, the inline `${{ fromJSON(steps.secrets.outputs.vault).X }}` in the three signing steps' `env:` blocks raised an opaque `The template is not valid ... Error reading JToken` error that masked the real cause.

## What this PR does

- **Hardens the three signing steps** (`GPG-sign binary`, `Code-sign and notarize macOS binaries`, `Re-GPG-sign macOS binaries`) to read secrets via `jq` from a `VAULT_SECRETS` env var instead of inlining `fromJSON`. On an empty/denied Vault response they now fail with an actionable `::error::` that names the role and the missing paths and points to the doc — instead of a template error. Signing behavior is **unchanged** when secrets are present.
- **Adds `docs/RELEASE-SIGNING-SETUP.md`** — the Vault access request for the RE/build-infra team: the role name, the two required paths (`cloudflare/warp-github-runner` and the unverified `sign/sonar-migration-tool` Apple-cert path), the evidence, the already-working paths, and re-trigger instructions.

## Follow-up required (not in this PR)

This PR makes the failure legible; it does **not** unblock releases. RE/build-infra still needs to grant the Vault role read access to the two paths documented in `docs/RELEASE-SIGNING-SETUP.md`. Once granted, push a `v*` tag (an `-rc` tag first is recommended) to validate end-to-end.

## Verification

- YAML validated (parses; all four jobs intact).
- Adversarial review (2 independent reviewers): variable flow correct both directions; `set -e` abort semantics preserved via the safe separate-`export` pattern; `jq`/here-strings/`base64 --decode` valid on the target runners; codesign script byte-for-byte unchanged after the insertion point.
- Secret scan: `gitleaks detect` (full tree) and `gitleaks protect --staged` both clean. `VAULT_SECRETS` is never printed; no `set -x`; every reference is a name/path/role, never a value.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,11 +100,23 @@ jobs:
             development/kv/data/sign key | GPG_SIGNING_KEY;
             development/kv/data/sign passphrase | GPG_SIGNING_PASSPHRASE;
 
+      # Read secrets from the raw Vault JSON via jq inside the script rather than
+      # inlining `fromJSON(...)` in `env:`. When Vault access is denied the JSON is
+      # empty, and an inline `fromJSON('')` raises an opaque "template is not valid"
+      # error; this guard fails with an actionable message instead.
       - name: GPG-sign binary
         env:
-          GPG_SIGNING_KEY: ${{ fromJSON(steps.secrets.outputs.vault).GPG_SIGNING_KEY }}
-          GPG_SIGNING_PASSPHRASE: ${{ fromJSON(steps.secrets.outputs.vault).GPG_SIGNING_PASSPHRASE }}
-        run: .github/scripts/gpg-sign.sh sonar-migration-tool-${{ matrix.suffix }}
+          VAULT_SECRETS: ${{ steps.secrets.outputs.vault }}
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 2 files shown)


okorach-sonar (approved):
LGTM

PR #469, opened 2026-07-03
PREQ-7041: align release pipeline to branch-based signing

tomverin:
## Summary

Aligns the full release pipeline with the standard SonarSource branch-based model (sonarqube-cli, sonar-dotnet-*, Azure Artifact Signing).

- **Removed** `v*` tag trigger — signing OIDC uses `refs/heads/main` and `refs/heads/branch-*`, not tags
- **Build + sign** on `main` and `branch-*` pushes (not PRs)
- **Publish** GitHub Release on `main` only; auto-creates tag `vYYYY.MM.DD-<run_number>`
- **Added** `sign-windows` job: Authenticode via `gh-action_azure-artifact-signing@v1`, then re-GPG-sign

## Signing matrix

| Platform | Mechanism |
|----------|-----------|
| Linux | GPG (`development/kv/data/sign`) |
| macOS | Apple codesign + notarize + stapler (`sign/apple`) + re-GPG |
| Windows | Azure Artifact Signing + re-GPG |

## Infra prerequisite (not in this PR)

`sign-windows` requires Azure onboarding in `re-service-config/azure_artifact_signing`:

- [BUILD-11815](https://sonarsource.atlassian.net/browse/BUILD-11815) — pool B release identity
- `sonar-migration-tool` module with `release_branches = ["main", "branch-*"]`

Until applied, `sign-windows` will fail OIDC; other jobs unaffected on `branch-*` if we skip windows or merge after infra.

## Test plan

- [ ] PR: test job only
- [ ] Merge + push to `branch-*`: build + sign jobs run, no GitHub Release
- [ ] Merge to `main`: full pipeline including publish (after Azure onboarding for Windows)

[BUILD-11815]: https://sonarsource.atlassian.net/browse/BUILD-11815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #471, opened 2026-07-03
fix(ci): fix workflow parse failure — env context in job-level if

okorach-sonar:
## Summary

- **Root cause:** PR #469 introduced `env.IS_RELEASE_BRANCH` and `env.IS_MAIN_RELEASE` in job-level `if:` conditions. GitHub Actions does not support the `env` context at the job level (only at step level, after the runner is instantiated) — this causes an immediate workflow parse failure: 0-second runs with "This run likely failed because of a workflow file issue."
- **Secondary issue:** The same PR changed runners to `sonar-xs`, which is private-repo-only. This repo is public, so those jobs would never start even if the parse issue were fixed - Currently we'll keep `sonar-xs` to see if this works, but we may switch to `github-ubuntu-latest-s` if it does not
- **Fix:** Replace `if: env.IS_RELEASE_BRANCH == 'true'` / `if: env.IS_MAIN_RELEASE == 'true'` with the actual `github` context expressions inlined directly. Kept `sonar-xs` to run (we may switch to `github-ubuntu-latest-s` if that does not work (consistent with fc4b173 which previously fixed this for the test job).

## Test plan

- [ ] Verify the CI run on this PR completes without a 0s "workflow file issue" failure
- [ ] Confirm the `test` job runs successfully on `github-ubuntu-latest-s`
- [ ] `build-binaries`, `sign-macos`, `sign-windows`, `publish` jobs are skipped on a PR (correct — they only run on `main` / `branch-*` pushes)
- [ ] After merge to `main`, confirm a full build-and-sign pipeline runs end-to-end

----
## Summary by Gitar

- **CI workflow fixes:**
  - Removed unused `IS_RELEASE_BRANCH` and `IS_MAIN_RELEASE` variables from `build.yml` env context.
- **Windows signing:**
  - Added `sign-windows` job to handle Authenticode signing via Azure and re-GPG-sign binaries.
- **Release workflow:**
  - Updated `publish` job to include `sign-windows` as a dependency and use a dynamic date-based release tag.
  - Added `branch-*` support to workflow `push` triggers.

<sub>This will update automatically on new commits.</sub>


kiloecho-sonar (dismissed):
- The changes of this PR fixed the primary issue, however the variables IS_RELEASE_BRANCH and IS_MAIN_RELEASE are now dead (lines 15–16).  They need to be removed in a later fix.  **We should fix this before merge to clean it up**
- The Test job runs on sonar-xs and passed in ~9 min on this public repo. So sonar-xs demonstrably works here — the premise of the "secondary issue" is wrong.
- Building binaries - all four show "skipping" on the PR run.
- Last item needs to be checked after merge to see if the full build-and-sign pipeline runs end-to-end.  **I'll add this as a task**

kiloecho-sonar (approved):
Removed unused variables IS_RELEASE_BRANCH and IS_MAIN_RELEASE.  Ready to merge

Issue #474, opened 2026-07-07
Add logic for handling unsupported programming languages.


daniel-borek-sonarsource:
When a transfer is triggered for a project that uses a 3rd party plugin for unsupported programming language, the issues and branches don't get transferred.  This is because the analysis endpoint fails with message like this:

Report contains a file with language 'name_of_the_programmning_language' but no matching quality profile...

The project still ends up in the cloud with correct permissions and quality gates, but no issues and no branches. 

It would be great to add some logic to detect such projects and display a meaningful explanation of what happened and why.  Perhaps we should add an option to not transfer such projects as well?

joshua-quek-sonarsource:
Fixed in **https://github.com/SonarSource/sonar-migration-tool/pull/490**.

### Why it happened

The fabricated scanner report describes languages twice, and the Compute Engine cross-checks them: every FILE component is stamped with the language the **source** server reported (`api/measures/component_tree`), while `metadata.qprofiles_per_language` can only name quality profiles that exist on the **target** organization. SonarQube Cloud has no analyzer — and therefore no quality profile — for a language contributed by a 3rd-party Server plugin, so the two disagree and the CE rejects the **entire** report, not just the offending file.

The project, its permissions and its quality gate are created in earlier phases, which is exactly why they survived while the code, issues and every branch did not. The non-main branches were skipped as a knock-on effect: the main-branch import is a blocking gate.

Three things kept it quiet: the rejection was logged at WARN, `importProjectData` tracked no per-project outcome (so the task summary said nothing), and the report rendered it as `API error when migrating project data: <raw CE dump>`.

### What changed

Before submitting, `transfer` now compares the languages of the report's file components against the target org's quality-profile languages — the CE's own check — and prints the languages, the file count, an example path, the cause, the consequence and the remedy. A new `--unsupported_languages` flag (validated before the extract phase) chooses the handling:

- **`exclude` (default)** — drop only the offending files, so every other file, issue, measure and branch still migrates. Reported as a Partial Migration naming the languages and file count.
- **`skip`** — submit no report for the project at all. This is the opt-out you asked about.
- **`warn`** — submit unchanged (previous behaviour), but the rejection is now explained rather than framed as an API error.

A rejected report is also no longer silent: it is logged at **ERROR** with the language and remedy, and counted, so the task summary reports `failed=1`.

[...]

PR #489, opened 2026-07-27
fix: migrate custom issue tags — stop dropping cross-language analyzer findings from the scan report (#456)

### LINKED ISSUE #456: Issue sync does not migrate issue custom tags, at least not for all issues
A project that had 3 issues for which a custom tag was set (namely sample project `demo-rules` with 3 issues with custom tag `action-plan`) did not see the same issues having the same tag after migration.
Worth noting, the 3 issues are in the regular OPEN status (a status that does not require status sync).

joshua-quek-sonarsource:
## The bug

Custom issue tags set on the source server were missing from the migrated issues. Reported against sample project `demo-rules`, whose three **OPEN** issues carrying the custom tag `action-plan` came out untagged on the target.

`Closes #456`

## Root cause

The issue's own hypothesis — that tag sync only runs for issues that also need a status transition — turned out to be **wrong**. `hasManualChanges` already treats a user-added tag as a standalone trigger, and the live run confirms it fired for exactly the right issues:

```
syncIssueMetadata: syncing pairs  source_total=7 actionable=3 custom_tags=3 accepted_or_false_positive=0
```

The real fault is one stage upstream: **the three issues never reached the target at all**, so there was nothing to tag.

`buildBranchReport` builds the fabricated scanner report in this order:

1. derive the report's language set from the language SonarQube assigned to each **file component**;
2. `filterRulesByLanguage` — filter the active-rule set to those languages;
3. `dropIssuesWithInactiveRules` — drop every native issue whose rule is not among the survivors (an orphan rule aborts the whole report in the Compute Engine, so this guard is necessary).

Cross-language analyzers break step 1's assumption. **Secret detection owns language `secrets`, but raises its issues inside files of other languages** — `demo-rules` has secrets findings in `secrets/az.xml` (language `xml`), `secrets/run_linters.sh` and `secrets/tmp` (SonarQube assigns these **no language at all**). So `secrets` never appeared in the component-derived set, all 359 secrets active rules were filtered out, and all 7 secrets issues were then dropped:

```
WARN dropped native issues referencing inactive rules  project=demo-rules-before-m456 dropped=7
INFO report packaged ... components=5 issues=1 externalIssues=0 activeRules=461
```

`issues=1` — and that one is the migrated Security Hotspot (tagged `former-hotspot`), not a native issue. `syncIssueMetadata` then correctly identified all 3 tagged issues as actionable and logged `no cloud counterpart on source line` for each, at **DEBUG**. That silence is why the loss looked like "tags didn't migrate" rather than "issues didn't migrate".

## Changes

**`go/internal/migrate/tasks_projectdata.go`**
- New `widenLangsForFindingRules()`: before the language filter runs, add to the report's language set the languages of the active rules the findings actually reference (native issues + hotspots). Gated on the **target organization exposing a quality profile for that language** — without one, the rules cannot be remapped to a valid qprofile key and the report metadata would cite a SonarQube Server profile key the CE rejects. Findings whose language has no target profile keep being dropped, with the existing warning.
- New `logWidenedLangs()` helper so the call site stays a single statement (avoids raising `buildBranchReport`'s cognitive complexity).

**`go/internal/migrate/tasks_issuesync.go`**
- `matchableIssue` gains `AllTags` — the issue's **complete** source tag list. `Tags` keeps its existing meaning (user-added subset, the actionability signal).
- `syncIssueTags` now writes `mergeIssueTags(sourceAllTags, cloudTags)`: the de-duplicated union of the source tag list and the tags the cloud issue already carries, plus the `metadata-synchronized` marker. `/api/issues/set_tags` **REPLACES** the tag list, so the previous behaviour of sending only the user-added subset stripped the rule-default tags the cloud analysis had assigned (e.g. `cwe`).
- Unmatched triaged issues are now reported at **WARN with totals** instead of only per-issue at DEBUG.

**`docs/`** — `MIGRATION-FACETS.md` (the Issue Tags row claimed "fully accurate"; corrected, and the Active Rules row now documents the language-filter trap), `TRANSFER-INTERNALS.md` (ASCII pipeline + two new "key facts"), `MIGRATION-CAPABILITIES.md` (SPEC-008).

SonarQube Server **2026.3.1** (`http://localhost:9000`) → SonarQube Cloud staging, org `open-digital-society-1`. Both runs held a global mutex so the three concurrent agents could not corrupt each other's org-level profiles.

Source state (unchanged across both runs):

```
$ curl -u $T: "http://localhost:9000/api/issues/search?componentKeys=demo-rules&tags=action-plan"
total=3
7990e7e8-4d6f-48b9-8e60-00c9c4e802e4  secrets:S7001                 secrets/az.xml         L7   OPEN  ['action-plan','cwe']
f8702c46-d9d1-430a-973e-344880fbe005  secrets:S6292                 secrets/run_linters.sh L7   OPEN  ['action-plan','cwe']
08c29033-c30b-4e1e-b974-4d3011a22459  secrets:My_custom_secret_rule secrets/run_linters.sh L16  OPEN  ['action-plan','cwe']
```

### BEFORE — `origin/main` build (381b068), target `demo-rules-before-m456`

```
WARN dropped native issues referencing inactive rules  dropped=7
INFO report packaged ... issues=1 activeRules=461
INFO syncIssueMetadata: syncing pairs  source_total=7 actionable=3 custom_tags=3
DEBUG no cloud counterpart on source line  rule=secrets:S7001 file=secrets/az.xml line=7
DEBUG no cloud counterpart on source line  rule=secrets:S6292 file=secrets/run_linters.sh line=7
DEBUG no cloud counterpart on source line  rule=secrets:My_custom_secret_rule file=secrets/run_linters.sh line=16
```

Target readback — **0 of 7 source issues migrated**, tag absent:

```
TARGET_TOTAL_ISSUES=1
TARGET_TAG_HISTOGRAM={"cwe":1,"former-hotspot":1,"secret":1}
AZ-j8l9H0wiZ1XcMwQaj | python:S2068 | secrets/hello_world.py | 3 | OPEN | ['cwe','former-hotspot','secret']

ACTION_PLAN_TOTAL=0        <-- the bug
```

### AFTER — this branch, target `demo-rules-after-m456`

```
INFO widened report language set for cross-language analyzer findings  languages=[secrets]
INFO report packaged ... issues=8 activeRules=820          (was issues=1 activeRules=461)
     (no "dropped native issues" warning)
INFO syncIssueMetadata: syncing pairs  source_total=7 actionable=3 custom_tags=3
WARN triaged source issues without a unique Cloud counterpart  actionable=3 synced=2 not_found=1
```

Target readback — **all 7 source issues migrated; the tag is present**:

```
TARGET_TOTAL_ISSUES=7
TARGET_TAG_HISTOGRAM={"cwe":7,"secret":7,"former-hotspot":1,"action-plan":2,"metadata-synchronized":2}

AZ-j_-Tz0wiZ1XcMwQj_ | secrets:S7001 | secrets/az.xml         | 7  | OPEN | ['action-plan','cwe','metadata-synchronized','secret']
AZ-j_-WF0wiZ1XcMwQkD | secrets:S6292 | secrets/run_linters.sh | 7  | OPEN | ['action-plan','cwe','metadata-synchronized','secret']
AZ-j_-V80wiZ1XcMwQkA | secrets:S7001 | secrets/hello_world.py | 10 | OPEN | ['cwe','secret']
AZ-j_-V80wiZ1XcMwQkB | secrets:S6292 | secrets/hello_world.py | 22 | OPEN | ['cwe','secret']
AZ-j_-WS0wiZ1XcMwQkE | secrets:S6292 | secrets/tmp            | 3  | OPEN | ['cwe','secret']
AZ-j_-WS0wiZ1XcMwQkF | secrets:S7001 | secrets/tmp            | 11 | OPEN | ['cwe','secret']
AZ-j_-V80wiZ1XcMwQkC | python:S2068  | secrets/hello_world.py | 3  | OPEN | ['cwe','former-hotspot','secret']

ACTION_PLAN_TOTAL=2        <-- was 0
```

Transfer exit code `0`; the CE `REPORT` task for `demo-rules-after-m456` finished `SUCCESS` with `warningCount=0`. Note the `cwe` rule-default tag is retained alongside `action-plan`, which is the `mergeIssueTags` union working — the pre-fix `set_tags` payload would have replaced it.

## Honest scope limit: 2 of 3, and why the third cannot be fixed here

The third tagged issue (`08c29033…`, rule `secrets:My_custom_secret_rule`, `run_linters.sh:16`) still does **not** migrate. That rule is a **template-instantiated custom rule that does not exist in the target organization**:

```
$ GET /api/rules/show?organization=open-digital-society-1&key=secrets:My_custom_secret_rule
HTTP 404
```

`restoreProfiles` created the target's `secrets` profile with 353 of the source's 359 rules; the template-instantiated ones were not created. With no such rule on Cloud the CE cannot raise the issue, so there is no target issue to tag — no change to the tag-sync path can recover it. This is a **separate, pre-existing gap in custom/template rule migration**, now documented under Known Gaps in `MIGRATION-FACETS.md` and surfaced at WARN as `not_found=1` rather than swallowed at DEBUG. Worth its own issue.

Also confirmed: including that non-existent rule in `activerules.pb` does **not** abort the report — the CE drops just that one issue and the task still returns `SUCCESS`.

## Interaction with #474

The agent fixing **#474** is changing the same language / active-rule / quality-profile reconciliation path from the opposite direction: its root cause is that the report stamps every FILE component with its source language while `metadata.qprofiles_per_language` can only carry languages that have a target profile, and the CE rejects the **entire** report on mismatch. It adds `go/internal/migrate/unsupported_languages.go` and touches `migrate.go` (WIP commit `b69dd00`, not yet merged).

- At the time of writing, **#474 has not landed** — `origin/main` is still `381b068`, this branch's base, so there is no textual conflict and my live verification is against current `main`.
- The two fixes are complementary in intent and, importantly, **agree on the gating condition**: my widening only admits a language when `scProfileByLang` has a target profile for it, which is exactly the constraint #474 is enforcing. My change never widens into a language that lacks a target profile, so it cannot reintroduce the `qprofiles_per_language` mismatch #474 fixes.
- **This has not been verified with both fixes applied together.** Whichever PR lands second should re-run the *live* transfer, not just the unit tests — a semantic disagreement about which languages/rules are admitted would not surface as a merge conflict. Happy to rebase and re-verify if #474 merges first.

## Tests

`make build`, `make test` (full suite), and `go vet ./...` all pass. New tests:

- `TestWidenLangsForFindingRules` — 8 cases: widening on a secrets finding in a non-secrets file; idempotence across many findings; **not** widened when the target has no profile for the language; language already present; rule absent from the active-rule list; active rule with no language; hotspot-driven widening; no findings.
- `TestWidenLangsForFindingRulesDegenerateInputs` — nil language map, empty rule list, input map left untouched.
- `TestSecretsFindingSurvivesLanguageFilter` — the end-to-end regression test: replays the real pipeline order (widen → `filterRulesByLanguage` → `dropIssuesWithInactiveRules`) over the `demo-rules` shape, asserts 2 issues are dropped **without** the fix and 0 **with** it, and that the report's `qprofiles` gains the `secrets` profile.
- `TestMergeIssueTags` — 8 cases: union with cloud rule-defaults, custom tag onto an untagged issue, cloud-only tags preserved, duplicate collapse, marker not duplicated on re-run, blank/whitespace tags dropped, marker-only, trimming.
- `TestMatchableIssueTagsToApply` — full list preferred, fallback to the user-only subset, empty.
- `TestOpenIssueWithCustomTagIsActionableAndNeedsNoTransition` — locks down the property the issue suspected was broken (an OPEN issue with only a custom tag is actionable and needs no transition), so it cannot regress.

### DIFF
--- a/docs/MIGRATION-CAPABILITIES.md
+++ b/docs/MIGRATION-CAPABILITIES.md
@@ -110,15 +110,24 @@ Automatic handling of projects with more than 10,000 issues, which hit SonarQube
 - Deduplication of results across overlapping date windows
 
 #### Issue Metadata Synchronization (SPEC-008)
+<!-- updated: 2026-07-27_22:37:13 -->
 
 After issues are uploaded via scanner reports, this phase matches each SonarQube Cloud issue back to its Server counterpart and replicates the original lifecycle metadata.
 
 **What it migrates:**
 - **Issue statuses**: OPEN, CONFIRMED, FALSE_POSITIVE, WONTFIX, ACCEPTED, RESOLVED, CLOSED
 - **Issue comments**: All comments with original author attribution (`[Migrated from SonarQube Server - @author]`)
-- **Issue tags**: Custom tags applied by users
+- **Issue tags**: Custom tags applied by users, written as the union of the source issue's full tag list and the tags the Cloud issue already carries (`set_tags` replaces rather than merges, so the union is what preserves both sides)
 - **Issue assignments**: User assignments mapped via `users.csv`
-- Pre-filtering skips issues with no manual changes (60-80% of typical enterprise issues), dramatically reducing API calls
+- Pre-filtering skips issues with no manual changes (60-80% of typical enterprise issues), dramatically reducing API calls. A user-added tag is itself a trigger, so an **OPEN** issue that only carries a custom tag is still synced — no status change is required.
+
+**Prerequisite — the issue must exist on the target.** This phase can only write onto a Cloud counterpart that the replayed scanner report actually produced. If an issue was dropped at report-build time (its rule was not in the report's active-rule set) or by the Compute Engine (its rule does not exist in the target organization), there is nothing to tag or transition. Such issues are counted as `not_found` and reported at WARN, e.g.:
+
+```
+WARN syncIssueMetadata: triaged source issues without a unique Cloud counterpart —
+     their status, comments and tags were NOT migrated
+     project=... actionable=3 synced=2 ambiguous_line=0 not_found=1
+```
 
 #### Hotspot Metadata Synchronization (SPEC-009)
 
@@ -339,7 +348,7 @@ Electron-based desktop application wrapping the CLI and browser GUI.
 | **External Issues** | **Not migrated** | **Full migration with ad-hoc rules** |
 | Issue Statuses | N/A | Synced (OPEN, CONFIRMED, FP, WONTFIX, etc.) |
 | Issue Comments | N/A | Synced with author attribution |
-| Issue Tags | N/A | Synced |
+| Issue Tags | N/A | Synced (full source tag list ∪ existing Cloud tags) |
 | Issue Assignments | N/A | Mapped via users.csv |
 | Hotspot Review Status | N/A | Synced (TO_REVIEW, REVIEWED/SAFE, etc.) |
 | Hotspot Comments | N/A | Synced |
--- a/docs/MIGRATION-FACETS.md
+++ b/docs/MIGRATION-FACETS.md
@@ -4,7 +4,7 @@
 > **Reconciled against the actual Go code on 2026-06-05.** Every row below reflects what the running binary does (`go/`, `lib/sq-api-go/`), not spec/design intent. Claims that previously described unbuilt specs or dead code have been corrected or moved. Full claim-by-claim evidence (file:line) is in [MIGRATION-FACETS-AUDIT.md](MIGRATION-FACETS-AUDIT.md).
 
 ## ✅ What IS Migrated
-<!-- updated: 2026-06-08_23:03:54 -->
+<!-- updated: 2026-07-27_22:37:13 -->
 
 The **Caveats / NOT carried** column is load-bearing — read it before relying on a row.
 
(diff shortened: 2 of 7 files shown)


joshua-quek-sonarsource:
Hi @okorach-sonar - Live E2E verification was done too to ensure that it really worked and the bug was squashed

okorach-sonar (approved):
LGTM

PR #480, opened 2026-07-08
fix: generate mandatory analytics.pb in fabricated scanner report

okorach-sonar:
## Summary
- Compute Engine now hard-rejects fabricated scanner reports missing `analytics.pb` (server-side change, RP-1034, effective 2026-07-06)
- Added the `Analytics` message to `scanner-report.proto` (mirrors the `ScannerReport.Analytics` schema from sonar-scanner-engine) and regenerated the Go bindings
- `PackageReport()` now writes `analytics.pb` with a single property `ci_name=sonar-migration-tool`, so migrated analyses are accepted by CE and identifiable as coming from this tool

Fixes #478

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/scanreport/...` (updated `TestPackageReportMinimal` to assert `analytics.pb` is present in the zip)
- [x] Manual: run a migration and confirm CE accepts the report (no "mandatory file 'analytics.pb' is missing" error)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #490, opened 2026-07-27
fix: handle programming languages SonarQube Cloud does not support (#474)

### LINKED ISSUE #474: Add logic for handling unsupported programming languages.

joshua-quek-sonarsource:
## The gap

Transferring a project that was analyzed with a **3rd-party (non-SonarSource) language plugin** produced a project on SonarQube Cloud with the right permissions, settings and quality gate — but with **no issues, no code and no branches**. The transfer printed `Transfer complete.` and exited `0`.

## Root cause

The fabricated scanner report describes languages in two places, and the Compute Engine cross-checks them:

| | source of truth |
|---|---|
| every FILE component's `language` | the **source** server (`api/measures/component_tree`) |
| `metadata.qprofiles_per_language` | quality profiles that exist on the **target** org (`buildSCProfileMap`) |

SonarQube Cloud has no analyzer — and therefore no quality profile — for a language contributed by a 3rd-party Server plugin. The two sets disagree and the CE rejects the **entire** report:

```
Report contains a file with language 'lua' but no matching quality profile
  (Visit failed for Component {key=...:src/main/java/demo/security/servlet/FileServlet.java,type=FILE} ...)
```

The project, permissions and gate are created in earlier phases, so they survive. Everything the report carries does not. Because the main-branch import is a blocking gate, the non-main branches were skipped as well.

Three things then hid it:
1. the rejection was logged at **WARN**;
2. `importProjectData` tracked no per-project outcome, so the task summary said nothing;
3. the report rendered it as `API error when migrating project data: <raw CE dump>` — no cause, no remedy.

## Changes

| File | Change |
|---|---|
| `go/internal/migrate/unsupported_languages.go` **(new, 390 L)** | `detectUnsupportedLanguages` compares the report's component languages against the target org's quality-profile languages — exactly the CE's own check. `applyUnsupportedLanguagePolicy` explains the finding and applies the mode. Also holds the CE-error classifier and the ERROR-level failure log. |
| `go/internal/migrate/tasks_projectdata.go` | One decision point in `buildBranchReport` (before `collectProjectLanguages`); `importResult`/`branchReport` carry the languages and excluded-file count; `recordImportOutcome` logs at ERROR and counts failures. |
| `go/cmd/transfer.go` | `--unsupported_languages`, validated **before** the extract phase so a typo'd mode fails fast instead of silently defaulting. |
| `go/internal/migrate/migrate.go`, `config_file.go` | `UnsupportedLanguages` on `MigrateConfig`/`Executor`, normalised in `applyDefaults`, readable from all four config shapes. |
| `go/internal/report/summary/collect.go` | `projectDataFailureReason` names the language, cause and remedy instead of "API error". |
| `go/internal/report/summary/project_failures.go` | `collectUnsupportedLanguageExclusions` routes a project whose files were excluded to **Partial** with an Issues line, so it is never reported as a full migration. |

### `--unsupported_languages`

| Mode | Behaviour |
|---|---|
| `exclude` **(default)** | Drops only the offending file components. Every other file, issue, measure and branch migrates. Reported as Partial, naming the languages and file count. |
| `skip` | Submits no report for the project at all — the *"option to not transfer such projects"* the issue asks for. |
| `warn` | Submits the report unchanged (previous behaviour), but the rejection is now explained. |

Excluding components is sufficient to drop everything attached to them — issues, measures, source, SCM blame and syntax highlighting all resolve through the component-ref map built from the surviving components.

**Safety guard:** `buildSCProfileMap` returns an *empty* map when the target-side profile lookup fails. Detection is disabled in that case rather than treating every language as unsupported, so a transient API error can never drop a project's whole file set. Unit-tested.

SonarQube Server **2026.3.1** → SonarQube Cloud **staging**, org `open-digital-society-1`, source project `checkstyle-issues` (2 branches; **237** issues / 10 hotspots / 449 ncloc / 14 files on `main`, **375** / 13 / 590 / 20 on `develop`).

No 3rd-party plugin is installed on the test server, so one was simulated at the API boundary: a reverse proxy in front of the source rewrote **one** file's `language` from `java` to `lua` in `api/measures/component_tree` (and added `lua` to `api/languages/list`). Verified surgical — same component count and `paging.total`, only the `language` field of `FileServlet.java` differs, all measures identical, other endpoints byte-identical. Everything downstream — report fabrication, submission, CE polling, the real CE rejection — is genuine.

### BEFORE — `origin/main` binary (`381b068`)

```
EXIT CODE: 0                                    ... "Transfer complete."
level=WARN msg="CE task failed" errorMessage="Report contains a file with
  language 'lua' but no matching quality profile ..."
level=WARN msg="project project data failed"
Task importProjectData: Duration 00:00:09.380   ... no failure count
```
Target `checkstyle-issues-b4-m474`:
```
branches: 1  -> master (empty, not even the source main branch name)
issues=0  hotspots=0  ncloc=-  files=-
CE tasks: 1 -> FAILED
```
Report: `Partial Migration | ... API error when migrating project data: CE task failed: ... Report contains a file with language 'lua' ...`
It then spent **8m30s** in `syncIssueMetadata` syncing 81 issues onto a project that had none.

### AFTER — this branch, default (`exclude`)

```
level=WARN msg="unsupported programming language: project \"checkstyle-issues-ex-m474\"
  branch \"main\" contains 1 file(s) in language(s) that have no quality profile on the
  target SonarQube Cloud organization: lua (1 file, e.g. src/main/java/demo/security/
  servlet/FileServlet.java). SonarQube Cloud only provides quality profiles for languages
  its own analyzers support, so a language contributed by a 3rd-party SonarQube Server
  plugin has none. The Compute Engine rejects an entire scanner report that contains a
  file whose language has no matching quality profile, which would leave this project on
  SonarQube Cloud with its settings, permissions and quality gate but with NO issues and
  NO branches. Excluding those files from the scanner report ..." languages=lua files=1 mode=exclude
level=WARN msg="excluded unsupported-language files ..." excludedFiles=1 remainingFiles=13
level=WARN ... branch="develop" ...                excludedFiles=1 remainingFiles=19
task summary task=importProjectData succeeded=1 failed=0
```
Target `checkstyle-issues-ex-m474`:
```
branches: 2
  - develop  main=False type=LONG  issues=160  ncloc=574  files=19
  - main     main=True  type=LONG  issues=226  ncloc=433  files=13
TOTAL issues: 386        CE tasks: 2 -> SUCCESS, SUCCESS
```
`main` went from 449→**433** ncloc and 14→**13** files — down by exactly the one excluded file (16 ncloc), which is the precise proof that only the `lua` file was dropped. Whole transfer: **1m16s** (vs 9m14s).

Report: `Partial Migration | ... Files in unsupported languages not migrated: lua (1 file) — the target organization has no quality profile for these languages — typically languages provided by 3rd-party SonarQube Server plugins. The files were excluded from the analysis report so the rest of the project could migrate; issues on them were not migrated`

### AFTER — `--unsupported_languages skip`

```
level=WARN msg="unsupported programming language: ... Skipping this project's data as
  requested by --unsupported_languages=skip: no scanner report is submitted for any of
  its branches. ..." branch=main  languages=lua files=1 mode=skip
... same for branch=develop
task summary task=importProjectData succeeded=1 failed=0
```
Target `checkstyle-issues-sk-m474`: **`CE tasks: 0`** — no report was ever submitted, versus 1 FAILED at baseline. Project, permissions and gate migrated; issues/branches deliberately not.

Report: `Partial Migration | ... Project data migration skipped: Project data not migrated: the project contains files in language(s) with no quality profile on the target organization — lua (1 file, e.g. src/main/java/demo/security/servlet/FileServlet.java) (typically a language provided by a 3rd-party SonarQube Server plugin). Skipped by --unsupported_languages=skip.`

## Tests

18 new unit tests (`go/internal/migrate/unsupported_languages_test.go`, `go/internal/report/summary/unsupported_languages_test.go`, `go/cmd/transfer_test.go`): mode parsing/validation/defaults, detection (language casing, empty-language files, per-language counts and sample paths), component exclusion, all three modes, the all-files-excluded edge case, the empty-profile-map safety guard, CE error classification, the ERROR log content, the report reason, and the Partial routing.

`make build`, `make test` (17 packages) and `go vet ./...` all pass. New code is at 100% function coverage except `detectUnsupportedLanguages` (95%), `applyUnsupportedLanguagePolicy` (93%) and `UnsupportedLanguageFromCEError` (88%). Cognitive complexity checked against the baseline: `runImportProjectData` was refactored back **below** its pre-change value so no new `go:S3776` violation is introduced; no new function or test exceeds 10.

## Interaction with #456

#456 (`fix/issue-456`, not yet merged) touches `buildBranchReport` from the other direction: `widenLangsForFindingRules` **adds** to the report's language set the languages of active rules the findings reference, so cross-language analyzers (e.g. `secrets`) are not stripped.

The two compose safely and there is no semantic conflict:

- This PR's decision point runs **before** `collectProjectLanguages`, subtracting file components whose language has no target profile. #456's widening runs **after**, adding rule languages.
- Both are gated on the **same** invariant — the target org must expose a quality profile for the language (`scProfileByLang`). #456 can therefore never re-add a language this PR removed, and this PR never removes a language #456 needs.
- A textual conflict is likely on the adjacent lines around `projectLangs := collectProjectLanguages(components)`. Correct resolution keeps the order: **#474 policy → `collectProjectLanguages` → #456 widen → `filterRulesByLanguage`**.

Whichever lands second should re-run its live verification. Happy to rebase onto #456 if it merges first.

## Honest limitations

- The unsupported language was simulated at the source API boundary (no 3rd-party plugin is installed on the test server, and installing one requires a restart that would have disrupted other agents sharing it). The CE rejection, the detection and all three modes were exercised against the real SonarQube Cloud CE.
- Target hotspots came back `0` in both the before and after runs. This is a **pre-existing** gap unrelated to this PR (also observed on `main` in the #480 review) — not a regression introduced here.
- Detection happens when the branch report is built, which is early enough to precede any submission for that project, but not a separate pre-flight before the extract phase.

Closes #474

### DIFF
--- a/docs/MIGRATION-CAPABILITIES.md
+++ b/docs/MIGRATION-CAPABILITIES.md
@@ -21,6 +21,25 @@ The following entities are fully migrated today:
 | **Portfolios** | Portfolio structure and project assignments |
 | **Project Data** | Optional protobuf report injection for historical analysis data |
 
+### Languages SonarQube Cloud does not support
+<!-- updated: 2026-07-27_23:16:00 -->
+
+SonarQube Server can analyze languages SonarQube Cloud cannot — anything
+contributed by a 3rd-party (non-SonarSource) plugin has no Cloud analyzer and
+therefore no quality profile there. Such files **cannot** be migrated: the
+Compute Engine rejects any analysis report containing a file whose language has
+no matching quality profile, and it rejects the report as a whole.
+
+`transfer` detects this before submitting and, by default, excludes only those
+files so that every other file, issue, measure and branch still migrates; the
+project is then reported as a **Partial Migration** naming the languages and the
+file count. `--unsupported_languages=skip` declines to migrate the project's
+data at all. See
+[TRANSFER.md](TRANSFER.md#unsupported-programming-languages---unsupported_languages)
+and the troubleshooting entry
+[Project migrated but has no issues and no branches](TROUBLESHOOTING.md#project-migrated-but-has-no-issues-and-no-branches).
+Issue #474.
+
 ---
 
 ## Planned Capabilities (Roadmap)
--- a/docs/TRANSFER-INTERNALS.md
+++ b/docs/TRANSFER-INTERNALS.md
@@ -234,6 +234,13 @@ statement of `runTransfer`.
   |         non-main branches SEQUENTIAL:                                      |
   |           buildBranchReport: load issues / hotspots / components /         |
   |             sources / activeRules ;  skip empty/purged ;                   |
+  |             buildSCProfileMap(org) -> profile per LANGUAGE ;               |
+  |             #474 applyUnsupportedLanguagePolicy(components, profiles):     |
+  |               langs(components) \ langs(profiles) = UNSUPPORTED            |
+  |                 exclude (default) -> drop those file components            |
+  |                 skip              -> return "skipped" + reason, NO submit  |
+  |                 warn              -> submit unchanged (CE will reject)     |
+  |               (empty profile map = API failure -> detection disabled)      |
   |             remap SQ->SC profile keys ;  dedup active rules ;              |
   |             drop issues on inactive rules ;                                |
   |             BackdateChangesets -> original creation dates                  |
@@ -474,7 +481,7 @@ flowchart TB
 ```
 
 ## Key facts the chart encodes
-<!-- updated: 2026-06-09_21:16:18 -->
+<!-- updated: 2026-07-27_23:15:00 -->
 
 - **Project scoping is carried by exactly one parameter.** `getProjects` is the
   sole consumer of `ProjectKeys`; it sets `projects=<key>` on
@@ -503,3 +510,19 @@ flowchart TB
   `GenerateReports` fails the transfer still prints "Transfer complete." and
   exits 0. Both report files are written to the **top-level** export dir, not
   inside `runDir`.
+- **The report's file languages and its quality-profile metadata come from two
+  different places, and the CE cross-checks them.** Component languages come
+  from the source server (`api/measures/component_tree`); `qprofiles_per_language`
+  can only name profiles that exist on the **target** org
+  (`buildSCProfileMap`). A language present in the former but absent from the
+  latter makes the CE reject the *whole* report with `Report contains a file
+  with language 'X' but no matching quality profile` — costing every issue and
+  every branch while the project, permissions and gate survive, because they
+  are created earlier. `applyUnsupportedLanguagePolicy` closes that gap by
+  comparing the two sets before submitting (#474). A failed profile fetch
+  yields an empty map, so detection is disabled in that case rather than
+  treating every language as unsupported.
+- **A rejected report no longer passes as success.** `importProjectData` logs
+  the failure at **ERROR** and counts it (`task summary … failed=1`); the
+  migration report names the offending language and the remedy instead of
+  framing it as "API error when migrating project data" (#474).
--- a/docs/TRANSFER.md
+++ b/docs/TRANSFER.md
@@ -1,5 +1,5 @@
 # Using `transfer` — Transfer One Project
-<!-- updated: 2026-06-05_10:50:51 -->
+<!-- updated: 2026-07-27_23:05:00 -->
 
 `transfer` is the **single-command**, **project-scoped** path. It chains the four phases of a migration — **extract → structure → mappings → migrate** — into one call, then writes a PDF summary on completion. Use it when you have one project (or a small, well-known set of projects) to move across.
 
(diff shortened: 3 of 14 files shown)


joshua-quek-sonarsource:
Hi @okorach-sonar , for unsupported languages, the implementation here is to do as such:
```
Partial Migration | ... Files in unsupported languages not migrated: lua (1 file) — the target organization has no quality profile for these languages — typically languages provided by 3rd-party SonarQube Server plugins. The files were excluded from the analysis report so the rest of the project could migrate; issues on them were not migrated
```

Also added a skip flag (you can read more above in claude's first comment in the PR)

okorach-sonar (approved):
LGTM

PR #504, opened 2026-08-10
fix: truncate user list to first 10 in permission limitation notes

### LINKED ISSUE #475: Skip the list of users/groups without any global permissions in the pdf report.
A project transfer will produce a pdf report that contains list of users for whom global permissions can't be transferred. The report lists all users, regardless of whether they have any permissions or not. 

In a case where we only have a small number of users and groups with any global permissions, but thousands of users overall, the report ends up being overly verbose, listing all usernames rather than only those that we should focus on.

okorach-sonar:
## Summary
- The Migration Limitations section of the PDF report listed every user with unmigrated global/permission-template SonarQube Server permissions, regardless of count. On instances with thousands of users, this made the report unreadably verbose even though usually only a handful of users hold those permissions.
- Added a formatLimitationUserList helper (10-user cap) and wired it into both the global-permissions and permission-template limitation messages in collectUserPermissionLimitations.
- <=10 affected users: unchanged behavior, all logins listed.
- >10 affected users: only the first 10 are listed explicitly, followed by a count of the rest, e.g. "first 10: u1, ..., u10 (and 2 more)".

Fixes #475

## Test plan
- [x] New table test for formatLimitationUserList covering empty / below-limit / exactly-at-limit / above-limit cases
- [x] New end-to-end CollectSummary test with 12 users confirming truncation and the remainder count, and that users beyond the first 10 are omitted
- [x] New end-to-end CollectSummary test with 4 users confirming all are still listed explicitly
- [x] go test ./... passes across the whole module
- [x] go build ./..., gofmt, go vet clean on the changed files
- [x] SonarQube Agentic Analysis (DEEP) run on both changed files; only pre-existing Cognitive Complexity findings unrelated to this change were reported

Generated with Claude Code

### DIFF
--- a/go/internal/report/summary/collect.go
+++ b/go/internal/report/summary/collect.go
@@ -239,6 +239,26 @@ func collectLimitations(runDir, exportDir string, mapping structure.ExtractMappi
 	return out
 }
 
+// maxListedLimitationUsers caps how many logins are spelled out in a
+// user-permission limitation note (#475) — instances with thousands
+// of users but only a handful holding the affected permissions were
+// producing an unreadably long PDF report.
+const maxListedLimitationUsers = 10
+
+// formatLimitationUserList renders logins for a user-permission
+// limitation note (#475): all of them when there are at most
+// maxListedLimitationUsers, otherwise only the first
+// maxListedLimitationUsers plus a count of the rest.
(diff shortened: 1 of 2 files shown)


kevin-lehen-sonarsource (approved):
LGTM

PR #491, opened 2026-07-27
feat(alm): migrate project DevOps platform bindings to SonarQube Cloud (#122)

joshua-quek-sonarsource:
## The gap

`sonar-migration-tool` never actually bound a migrated project to its DevOps platform repository on SonarQube Cloud.

The scaffolding existed (`matchProjectRepos` + `setProjectBinding` in `go/internal/migrate/tasks_alm.go`), but **three independent defects each made it a no-op against a live SonarQube Cloud instance**, and a fourth kept it out of the `transfer` path entirely:

1. **`project_id` was always empty.** `matchProjectRepos` read `id` from `getProjectIds`, which is the `components` array of `/api/projects/search`. SonarQube Cloud does not return an internal id there. `setProjectBinding` then hit `if projID == "" { return nil }` and skipped silently — no counter, no warning, no error.
   ```
   $ curl -s -u "$T:" 'https://sc-staging.io/api/projects/search?organization=open-digital-society-1&ps=1'
   {"components":[{"organization":"open-digital-society-1","key":"...","name":"...",
     "qualifier":"TRK","visibility":"private","lastAnalysisDate":"..."}]}   # <-- no "id"
   ```
2. **`repository_id` was always empty.** `MatchDevOpsPlatform` read `id` from each `/api/alm_integration/list_repositories` entry, but Cloud returns `{label, installationKey, slug, linkedProjects, private}` — no `id` — so nothing ever matched.
3. **Wrong host.** The write went to `e.Cloud` (`https://sonarcloud.io`). `/dop-translation/project-bindings` is served **only** by the enterprise host; the standard host returns the SPA index for that path.
4. **`transfer` excluded it.** `transferTargetTasks` omitted both ALM tasks, and `docs/TRANSFER.md` listed bindings under "Not modified".

The existing test suite passed throughout because its fixtures hard-coded `"id"` fields that the real API does not send, and the enterprise-API mock had a permissive catch-all route.

## The design

Three tasks, `getOrgBinding` → `matchProjectRepos` → `setProjectBinding`:

- **`getOrgBinding`** (new) reads each target organization's own DevOps binding via `GET /api/alm_integration/show_bound_organization`, inferring the platform from the ALM URL host. An unbound org is recorded as `bound:false` rather than omitted, so "not bound" is distinguishable from "not looked up". 400/403/404 are treated as unbound instead of failing the run.
[...]

Issue #122, opened 2026-05-20
`sonar-migration-tool` should migrate projects DevOps platform binding


okorach-sonar:
- [ ] When project is bound on the source SQS, the same binding should be migrated to the SQC platform. This only concerns the project level binding as the devops platform binding requires secrets that can't be migrated. Specifically:
       - GitLab: Project Id
       - GitHub: Repository name.
       - Azure DevOps: Project Name + Repository Name
       - BitBucket Cloud: Repository slug
- [ ] The project should be automatically migrated in the organization that is bound to the given DevOps platform organization the project belongs to
- [ ] The project binding should be attempted only if the project is bound on the SQS side (obvious but better write it)
- [ ] The project binding should be attempted only if the org is bound. If the SQS original project was bound but the target org is not bound, then in the migration report the project migration status should be set to "Partially Migrated" and the Details column should specify that "project binding was not possible because the org itself is not bound"
- [ ] Replicating the SQS project binding on SQC may require some adaptations as the binding process is different on SQC than it is on SQC
- [ ] If the project is bound on the SQS side but the org is not bound on the SQC side, the project migration report should report the project as "Partial migration" and a comment in the detail section should state that the project binding could not be migrated because the SQC org is not bound
- [ ] If the project is bound on the SQS side and the SQC is bound, but the project binding migration fails, the project migration report should report the project as "Partial migration" and a comment in the detail section should state that the project binding could not be migrated because of an error (and if possible provide the API error in the migration report)

joshua-quek-sonarsource:
PR: https://github.com/SonarSource/sonar-migration-tool/pull/491

## What was wrong

The project-level ALM binding scaffolding already existed (`matchProjectRepos` + `setProjectBinding`), but **three independent defects each made it a silent no-op against a live SonarQube Cloud instance**, and a fourth kept it out of `transfer` altogether:

1. `project_id` was read from `/api/projects/search`, which on Cloud returns **no** `id` → the write was skipped with no log, no counter and no error.
2. `repository_id` was read from `id` on `/api/alm_integration/list_repositories` entries, but Cloud returns `{label, installationKey, slug, ...}` — no `id` → nothing ever matched.
3. The write went to the standard host; `/dop-translation/project-bindings` is served **only** by the enterprise host (the standard host returns the SPA index).
4. `transferTargetTasks` omitted both ALM tasks, and `docs/TRANSFER.md` listed bindings under "Not modified".

[...]

PR #507, opened 2026-08-11
fix: migrate aborts when the target organization has no DevOps platform binding (#505)

### LINKED ISSUE #505: `migrate` error on `getOrgBinding`
The `migrate` command fails on recently added step `getOrgBinding`.
Below are the last few logs of the migrate command:

```
time=2026-08-10T14:34:48.393+02:00 level=WARN msg="retrying request" method=POST endpoint=/api/webhooks/create status=0 attempt=1 maxAttempts=4
time=2026-08-10T14:34:48.393+02:00 level=WARN msg="setGlobalWebhooks failed" org=latest-unbound name=Jenkins url=https://my.jenkins.server/sonar-webhook/ err="Post \"https://sc-staging.io/api/webhooks/create\": context canceled"
time=2026-08-10T14:34:48.393+02:00 level=INFO msg="task summary" task=setGlobalWebhooks succeeded=1 failed=3 total=4 duration=00:00:01.893
time=2026-08-10T14:34:48.399+02:00 level=INFO msg="Command migrate: Duration 00:00:02.527"
PDF summary report: migration-files-ee/migration_summary.pdf
Markdown summary report: migration-files-ee/migration_summary.md
Error: phase 2: task getOrgBinding: http: read on closed response body
```

This probably mean that the recently added feature of migrating project bindings (https://github.com/SonarSource/sonar-migration-tool/issues/122) does not work. This must be fixed

joshua-quek-sonarsource:
Fixes #505.

## Severity

Every `migrate` / `transfer` run into a SonarQube Cloud organization that is **not bound to a DevOps platform** aborts and creates **nothing**. The task that fails, `getOrgBinding`, was introduced by the #122 fix (#491), so this is a regression in that feature — and it fires on the common case, not an edge case: two of the three organizations available on our staging instance are unbound.

## Root cause — two defects in series

**1. SonarQube Cloud answers HTTP 500 for an unbound organization.** Probed live:

| Organization | `GET /api/alm_integration/show_bound_organization` |
| --- | --- |
| `open-digital-society-1` (bound to GitHub) | `200` `{"almOrganization":{"key":"Open-Digital-Society",...}}` |
| `teams-demo-org` (unbound) | **`500`** `{"errors":[{"msg":"An unexpected error occurred. Please try again later."}]}` |
| `smt-505-unbound` (unbound, created for this test) | **`500`** same body |
| `nonexistent-org-xyz` | `404` `Could not find organization with key '...'` |

It is reproducible per organization, so it is Cloud's normal answer for "no binding", not a transient fault. `404` means *no such organization* — the code comment claimed the opposite.

**2. The retrying transport destroyed the response.** `500` is retryable, so `retryTransport.RoundTrip` exhausted its schedule and then hit this:

```go
rl := t.observeRateLimit(resp)
drainAndClose(resp)                      // body drained AND closed
wait, total, more := t.nextWait(rl, attempt)
...
if !more {
    return resp, err                     // err == nil, body already closed
}
```

The caller's `io.ReadAll(resp.Body)` then failed with `http: read on closed response body`. The real status was gone, so `getOrgBinding`'s `IsHTTPError(err, 400, 403, 404)` check could never match, it returned a hard error, and the errgroup cancelled every sibling task mid-flight — the `context canceled` lines on `setGlobalWebhooks` in the report, and on `createGroups` in my reproduction.

## The fix

**`lib/sq-api-go/retry.go`** — decide, then dispose. `planRetry` computes whether another attempt will happen *before* touching the body: drain-and-close when retrying (connection reuse, unchanged), leave it **readable** when giving up. For 429s the bytes the classifier peeked at are spliced back with `io.MultiReader` behind a wrapper that still closes the original body. Observer, gate extension, recovery callback, retry log and the context-cancellation branch are behaviourally unchanged.

It also fixes a second real defect found while testing this: retries never rewound `req.Body`. With `GetBody` set and a **fresh** connection a retried POST failed outright (`ContentLength=N with Body length 0`); with `GetBody` nil it silently sent an **empty** body and got a 200 — a failed write turning into a different, apparently-successful call. Retries now rewind via `http.Request.GetBody`, and a request with a body but no `GetBody` is not retried at all rather than replayed as something else.

**`go/internal/migrate/`** — reading an organization's DevOps binding only enables the optional project-level binding, so no failure of it may abort a migration. `getOrgBinding` now records `400/403/404/500` as a plain unbound org (INFO) and anything else as unbound + `lookup_error` (WARN); `getOrgRepos` does the same for `list_repositories`, which answers `400 "This organization is not bound to an ALM application"`. A cancelled or timed-out run still propagates.

**Report honesty.** Saying "the org is not bound" when the lookup never succeeded asserts something the tool did not observe, so `org_binding_unknown` / `repos_unknown` get their own Details sentences plus the underlying API error — the error surfacing #122 asked for. A genuinely unbound org still reads *"project binding was not possible because the org itself is not bound"* and still marks the project Partial Migration, unchanged.

## Live verification

Source SonarQube Server 2026.3.1 @ `localhost:9000` → SonarQube Cloud staging `sc-staging.io`. **No fault injection** — the 500 is what Cloud really returns. Every run is the real binary against the real service.

Same command, three builds, against the real unbound org `smt-505-unbound`:

| Build | Outcome |
| --- | --- |
| `origin/main` | exit 1 — `Error: phase 2: task getOrgBinding: http: read on closed response body` |
| transport fix only | exit 1 — `Error: phase 2: task getOrgBinding: HTTP 500 GET .../show_bound_organization?organization=smt-505-unbound - An unexpected error occurred. Please try again later.` |
| **both fixes** | **exit 0** — `getOrgBinding: organization is not bound to a DevOps platform`, record `{"bound":false,...}` |

The middle row is the point: the transport fix alone restores the true status but the run still dies, so both halves are required.

A full `transfer` on `origin/main` reproduces the reporter's log verbatim, including the collateral damage and the empty target:

```
level=ERROR msg="task failed" task=getOrgBinding err="http: read on closed response body"
level=WARN  msg="createGroups: create failed" name=sonar-administrators err="Post \"...\": context canceled"
Error: migrate failed: phase 2: task getOrgBinding: http: read on closed response body
→ projects in smt-505-unbound: 0
```

**End-to-end, both binding outcomes, with the fix:**

*Source project bound, target org not bound to that platform* — `demo:gitlab-ci-maven` (GitLab-bound) → `open-digital-society-1` (GitHub-bound). Transfer exit 0, project created, 5 issues migrated, report:

```
| GitLab-CI / Maven | open-digital-society-1 | Partial Migration | New Project Key: demo:gitlab-ci-maven-m505
  100% of issues with manual changes synced (3/3)
  project binding was not possible because the org itself is not bound |
```

*Both bound — the #122 happy path, to prove no regression* — `alm-e2e` (GitHub-bound) → `open-digital-society-1`:

```
INFO setProjectBinding: DevOps platform binding created project=alm-e2e-m505b repository=Open-Digital-Society/sonarqube-example
     task summary task=setProjectBinding succeeded=1 failed=0 total=1
```

Confirmed independently on the target:

```
GET /api/alm_integration/is_project_bound?project=alm-e2e-m505b  -> {"is_bound":true}
GET {api}/dop-translation/project-bindings?projectId=AZ_vPC2A3JSKilK5A4oL
  -> {"bindings":[{"devOpsPlatform":"github","repositoryId":"625940442",
       "slug":"Open-Digital-Society/sonarqube-example",...}],"total":1}
```

An earlier attempt at that run returned `400 "This repository is already bound to another project"` because #122's test project still held the binding — which incidentally live-verified the other #122 requirement, the API error reaching the report: *"DevOps platform binding not migrated — sonarqube api error: POST .../dop-translation/project-bindings → 400: {...}"*.

## Quality gate

`gocognit` (== S3776) over the touched files, against `origin/main`: the four functions above the threshold are pre-existing and unchanged, and `(*retryTransport).RoundTrip` **drops off** the list (18 → 12). This branch removes one finding and adds none.

`go build`, `go vet`, `go test ./...` pass in both modules (`go/` — 17 packages — and `lib/sq-api-go/`, the latter also under `-race`).

## Also here: on-premise DevOps platforms (adjacent, separate commit)

SonarQube Cloud integrates only with **cloud** platforms — GitHub.com, GitLab.com, Azure DevOps Services, Bitbucket Cloud. `matchProjectRepos` already declined to bind anything else (the `!info.IsCloud` guard), so no wrong repository was ever bound — but it did so with a bare `continue`: no record, nothing in the report, and the project presented as **Full Migration** even though a real source binding did not come across.

Live, a project bound to **Bitbucket Server**:

```
before  | Project 3 | ... | Full Migration    | New Project Key: test:project3-m505e |
after   | Project 3 | ... | Partial Migration | New Project Key: test:project3-m505e
                                                project binding was not possible because the source project is bound to
                                                an on-premise DevOps platform, which SonarQube Cloud cannot integrate with |
```

A project with **no** source binding still writes nothing and stays Full Migration — that is the #122 behaviour; the old code conflated the two cases.

Second, `cloudEndpoints` gained `visualstudio.com`. Azure DevOps **Services** accounts created before the `dev.azure.com` rename still bind as `https://<account>.visualstudio.com` and are cloud, but were classified on-premise and silently dropped. The target-side `almHostPatterns` has always had it — the two lists had drifted. `IsCloudBinding` now also lowercases before matching, as `almFromURL` already does.

Note this **changes the reported status** of an on-premise-bound project from Full to Partial Migration. That is the honest outcome: something the operator had on the source is not on the target.

## Not fixed here

Organization-level binding still cannot be migrated. `POST /api/alm_integration/bind_organization` requires an `installationId` (GitHub / Bitbucket Cloud App) or a GitLab `almAccessToken` — DevOps-platform secrets that do not exist anywhere in a source SonarQube Server export. The target organization must still be bound by hand first. This confirms the limitation #122 recorded.


okorach-sonar (approved):
LGTM

PR #522, opened 2026-08-12
GUI checkboxes for project data migration and issue sync (#516)

### LINKED ISSUE #516: GUI should have 2 tick boxes for project data migration and issue sync
Project data migration and issue sync are 2 optional tasks of the migration tool, that were introduced after the current version of the GUI.
So:
- [ ] The GUI should have 2 tick boxes to select those tasks or not. The tick boxes should be on the 1st screen of the GUI, below the fields for source URL and source token
- [ ] The tick boxes should be enabled (ticked) by default, if no config file is given or if the setting is not specified in the config file
- [ ] If the settings are configured in the config file, the tick boxes should honor the config file value (true or false)
- [ ] If the project data migration is turned off, the issue sync tick box should be grayed (can't be changed) and turned off too 
- [ ] The settings for project data migration and issue sync set in the GUI should be honored in downstream migration steps

okorach-sonar:
## Summary
- Adds two checkboxes to the GUI wizard source-credentials screen (below URL/token) to opt out of project-data migration and/or issue sync, wiring the existing `skip_project_data_migration` / `skip_issue_sync` flags through the wizard instead of hardcoding "include everything".
- Checkboxes default to checked; when a `--config` file is passed to `gui`, they honor the file values via a new wizard-state seed field.
- Unchecking "project data migration" grays out and forces off "issue sync" in the UI, mirroring the existing backend cascade in `migrate.MigrateConfig`.
- Settings collected in the GUI now flow into `extract.ExtractConfig` / `migrate.MigrateConfig` for the downstream extract and migrate steps.

Closes #516

## Test plan
- [x] go build ./...
- [x] go vet ./...
- [x] go test ./...
- [ ] Manually run go run . gui and confirm the checkboxes render, default to checked, and issue-sync grays out when project-data is unchecked
- [ ] Manually run go run . gui --config config_example.json and confirm checkbox defaults follow the config file

Generated with Claude Code

### DIFF
--- a/go/cmd/gui.go
+++ b/go/cmd/gui.go
@@ -178,6 +178,14 @@ func resolveGUIDefaults(cmd *cobra.Command) (string, *wizard.WizardState, error)
 		v := migrateCfg.EnterpriseKey
 		seed.EnterpriseKey = &v
 	}
+	// #516: extractCfg and migrateCfg parse the same unified top-level
+	// skip_project_data_migration / skip_issue_sync config keys, so
+	// extractCfg is used as the single source for the wizard's
+	// migration-scope checkboxes.
+	includeProjectData := !extractCfg.SkipProjectDataMigration
+	seed.IncludeProjectData = &includeProjectData
+	includeIssueSync := !extractCfg.SkipIssueSync
+	seed.IncludeIssueSync = &includeIssueSync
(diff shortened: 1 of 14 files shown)


kevin-lehen-sonarsource (approved):
LGTM

Issue #520, opened 2026-08-12
`sonar-migration-tool` should compute a log an indicative overall progress status


okorach-sonar:
The `extract` and `mappings` commands can be rather long. To given users some idea of the progress and ETA, these commands should log an overall progress estimation and ETA at regular interval.
The progress bar should be mostly indexed on projects since this is expected to be the objects that are by far the most numerous to migrate.

Computing a progress is complicated given how the tool runs the migration. It does not handle each object (eg each project) one by one (dealling with all migration aspects of the objects before moving to the next). It rather iterates on all objects to execute 1 single migration aspect (eg create the project), then iterates on all objects for a 2nd migration aspect (eg set permissions), the a 3rd (eg configure projects QG) etc... 
We'll apply approximate estimations of the progress as below:
- [ ] All migration (extract or migrate) of objects except projects should be weighted as 5% of the total migration. The remaining 95% will be assigned to projects (extract or migrate)
- [ ] Out of the 95% assigned to project (extract or migrate) 20% will be assigned to project configuration (extract or migrate), 25% to project data (extract or migrate) and 50% to issue sync (extract or migrate)
- [ ] Every extract or migrate sub-task within the "All other objects", "project configuration", "project data" or "issue sync" should have the same equal "weight" (Example of tasks: getPlugins, getGates, getServerSettings, getRules, getGateConditions, getNewCodePeriods, getProjectSCMData, getProjectWebhookDeliveries, etc...)
- [ ] When project data migration and/or issue sync migration is turned off, the weight of these steps should be counted as 0 in the overall migration
- [ ] The tool should the compute an overall migration progress from the above split
- [ ] The overall formula is:
  - If everything turned on: <% of general tasks completed> * 5% + <% of project config tasks completed> * 20% + <% of project data tasks completed> * 25% + <% of project issue sync tasks completed> * 50% 
  - If issue sync is turned off: (<% of general tasks completed> * 5% + <% of project config tasks completed> * 20% + <% of project data tasks completed> * 20%) * 100 / 50
  - If project data migration is turned off: (<% of general tasks completed> * 5% + <% of project config tasks completed> * 20%) * 100 / 25
- [ ] The ETA should be computed from the time consumed by the command to far and the progress percentage.
- [ ] The tool should log progress and ETA every 30s. Like "Overall progress: 72% - ETA: hh:mm:ss"
[...]

PR #523, opened 2026-08-13
GUI progress bar with ETA for extract/migrate (#519)

### LINKED ISSUE #519: The `extract` and `migrate` commands should have a progress bar
The `extract` and `mappings` commands can be rather long. To given users some idea of the progress and ETA, the GUI should show a progress bar.
- [ ] The progress bar should indexed on the overall progress calculation described in #520
- [ ] Right to the progress bar a field should display ETA estimation, as calculated in #520. The ETA should be rounded to the nearest minute
- [ ] The progress bar should be below the spinner and the "Processing" message (See below screenshot)
Image

okorach-sonar:
## Summary
- Adds a live progress bar + ETA field to the GUI wizard below the spinner/"Processing..." message, closing #519. common.Tracker (#520) already computed a run-wide percent/ETA every tick but only logged it internally; this threads it out through a new ProgressCallback on ExtractConfig/MigrateConfig -> a new Prompter.DisplayOverallProgress method -> a new display_overall_progress WebSocket message, with no change to CLI behavior (the callback is nil for CLI callers).
- Tracker.Start now emits one snapshot immediately instead of waiting for the first tick, so short runs no longer show nothing until jumping straight to 100%.
- Tick interval reduced from 30s to 10s.
- ETA under 2 minutes is shown as exact seconds instead of "~1 min"/"~0 min"; at/above 2 minutes it's rounded to the nearest minute.
- Fixes a pre-existing bug found while testing the above: WizardState.Save used a non-atomic os.WriteFile, so a Ctrl-C/crash mid-save could leave .wizard_state.json at 0 bytes, breaking every subsequent gui run with "unexpected end of JSON input". Save now writes via temp-file + rename, and Load treats an already-corrupted empty file as absent state.
- Relabels the GUI header badge from "Experimental" to "Beta".

## Test plan
- [x] go build ./...
- [x] go vet ./...
- [x] go test ./...
- [x] Manually run go run . gui, start a wizard run, confirm the progress bar appears at 0% immediately and advances to 100%
- [x] Confirm CLI-only extract/migrate runs are unaffected (same log line, just every 10s now)
- [x] Confirm a previously-corrupted (0-byte) .wizard_state.json no longer breaks gui startup

Generated with Claude Code

### DIFF
--- a/go/internal/common/progress_tracker.go
+++ b/go/internal/common/progress_tracker.go
@@ -121,6 +121,8 @@ type Tracker struct {
 	mu        sync.Mutex
 	completed map[string]bool
 
+	onUpdate func(percent float64, eta time.Duration, known bool)
+
 	stopOnce sync.Once
 	stopCh   chan struct{}
 	doneCh   chan struct{}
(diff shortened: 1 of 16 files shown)


joshua-quek-sonarsource (dismissed):
tested and the latest changes pushed 10-20mins ago broke the build on my end. Trying to implement a fix.

joshua-quek-sonarsource (approved):

Issue #423, opened 2026-06-26
Adjust migration tool to the fact that SQC drops hotspots starting July 1st, 2026


okorach-sonar:
- [ ] When the target is SQC (all the time as of now), the `sonar-migration-tool` should migrate hotspots as issues, since hotspots can't be migrated anymore.
- [ ] The way the SMT should convert hotspots into issues should mimic the way the real scanner does it
- [ ] The hotspots converted as issues and migrated should have the `sqs-hotspot` tag assigned

joshua-quek-sonarsource:
PR: https://github.com/SonarSource/sonar-migration-tool/pull/492

## What I found

I probed the staging Cloud instance first rather than assuming. Former hotspot rules were **converted in place** into ordinary issue rules — `python:S4502` and `java:S2245` are now `VULNERABILITY`, `python:S1313` is `CODE_SMELL`, all carrying SonarSource's `former-hotspot` system tag. A few old ones (`python:S4823`, `python:S4784`) were **retired outright** (`status: REMOVED`), so findings on those are unmigratable.

One correction to the premise worth recording: a baseline transfer on `main` showed hotspots were **already arriving as ordinary issues** with the right rule, file, line and backdated date. That is structural — the native `Issue` protobuf has no type field, so the scanner only ever names a rule and the server classifies the finding. The "31 → 0" observation came from measuring the target with `/api/hotspots/search`, which now returns 0 by definition.

The real defects were: no marker identifying a migrated finding as a former hotspot; triage that **never synced at all**, because the target lookup went through `/api/hotspots/search` so every hotspot silently resolved to `not_found`; a fabricated severity override; an orphan-rule path that let a hotspot on a retired rule abort the entire report in the CE; and an extract-side `return nil` that discarded already-collected hotspots on a 403/404.

## The fix

Hotspots are converted to native issues that name the rule and stamp **no type and no impacts**, so Cloud classifies them exactly as it would for a real scan. Review state is applied afterwards as an issue transition (`TO_REVIEW`→OPEN, `SAFE`→FALSE_POSITIVE, `FIXED`/`ACKNOWLEDGED`→ACCEPTED — `ACKNOWLEDGED` is no longer downgraded to `SAFE`), and every migrated hotspot is tagged `sqs-hotspot` via `/api/issues/set_tags`. Converted hotspots now also pass the same active-rule filter as native issues, so a retired rule is dropped and counted instead of taking the branch down.

## Live verification

[...]

PR #531, opened 2026-08-19
Improve hotspot sync algorithm + add --fast_sync option (#527)

### LINKED ISSUE #527: Improve algorithm for hotspot sync to improve migration speed
Currently the migration tool looks at every source hotspots to eventually sync it with a target one.
This is not efficient because, statistically a lot of hotspots are likely to be in their original TO_REVIEW state, meaning they have not been reviewed whatsoever. These hotspots need no sync, and trying to find a match on the target SQC is time consuming.
So we want to change the algorithm to select hotspots to sync:

- [ ] All the hotspots extracted should be extracted with their state (TO_REVIEW, ACKNOWLEDGED, SAFE, FIXED), and comment history
- [ ] The hotspot sync phase source only be run on source hotspots that are not in TO_REVIEW state or that have (user created) comments. Hotspots that do not meet any of those 2 criterias should be excluded from the sync. User created comment means comments that are not technical comments that might have been created by SonarQube itself.
- [ ] Hotspots in ACKNOWLEDGED state should not be synced (unless they have user comments in which case the comment sync should be performed) but still inventoried for reporting purposed.
- [ ] The reporting on projects hotspots sync (% of hotspots synced) should be calculated on the hotspots that are eligible to sync (defined in criteria above) and hotspots in ACKNOWLEDGED state

[EDIT]
The reason for processing all hotspots is for back linking the SQC hotspot with the original SQS hotspot. This is however very costly and not so useful when the original hotspot has had 0 changes. So we're adding the following:

- [ ] The tool (`transfer` and `migrate`) should accept an option to turn off syncing hotspots (and issues) that are in original state with no user changes (original state - OPEN for issues, TO_REVIEW for hotspots -, no custom tags, no user comments)
- [ ] The option is `--fast_sync`. It can be used from the CLI or in the config file
- [ ] By default `fast_sync` shall be turned off. It can be turned on explicitly through CLI or config file
- [ ] The `--fast_sync` should be documented in the `transfer` and `migrate` command, as well as in the recap of all configuration options
- [ ] The migration report should document in the bottom remarks ("run metadata" section) whether the `fast_sync` option was used or not

okorach-sonar:
## Summary
- Hotspot sync now classifies each source hotspot as **excluded** (TO_REVIEW, no user comment), **acknowledged** (REVIEWED+ACKNOWLEDGED), or **eligible**, and replaces the previous per-hotspot Cloud search with a bulk per-(project, branch) index build (`buildCloudIssueIndex`).
- Every hotspot is still resolved and tagged `sqs-hotspot` from that cached index (#423), but only eligible/acknowledged hotspots get the state-transition + comment sync; the %-synced reporting denominator is now the eligible+acknowledged set instead of every hotspot.
- Extraction now fetches hotspot detail (comments) for every status, not just REVIEWED, so a TO_REVIEW hotspot with a genuine user comment is still detected as eligible.
- Real-world testing (743-hotspot project, none touched) showed the remaining per-hotspot tag + back-link write calls still dominate runtime even after the search was batched. Added `--fast_sync` (usable on `migrate`, `transfer`, and `sync-issues`, via CLI flag or config file — top-level or inside the `target` subsection, target wins) which skips those two writes for hotspots with zero user changes on the source (original TO_REVIEW state, no user comment, no custom tags). Defaults to **off** — current behavior (every hotspot tagged + back-linked) is unchanged unless explicitly opted in.
- Confirmed via code research that the issue-sync path already skips untouched issues unconditionally (`hasManualChanges`), so `--fast_sync` has no additional effect there today — documented in place rather than adding dead logic.

## Test plan
- [x] `go build ./...`, `go vet ./...`
- [x] `go test ./... -race` (full suite green)
- [x] New unit tests: hotspot eligibility classification, bulk index build/chunking, category-gated tag/back-link skipping under `--fast_sync`, config-file resolution (all 4 shapes + target-override) for `fast_sync`, CLI flag override tests for `migrate`/`transfer`/`sync-issues`
- [ ] Manual: re-run against the real `laravel` project (743 untouched hotspots) with `--fast_sync` and confirm the sync duration drops sharply from the observed 3.5 minutes

Closes #527

### DIFF
--- a/go/cmd/migrate.go
+++ b/go/cmd/migrate.go
@@ -67,6 +67,7 @@ func init() {
 	f.Bool("skip_profiles", false, "Skip quality profile migration/provisioning in SonarQube Cloud")
 	f.Bool(flagSkipIssueSync, false, "Skip the final per-issue and per-hotspot metadata sync (#299). Same semantics as the skip_issue_sync config-file field — defaults to false (sync happens); pass the flag to skip.")
 	f.Bool(flagSkipProjectDataMigration, false, "Skip the entire project-data migration: importProjectData and the trailing per-issue/per-hotspot sync (#303). Defaults to false (data is migrated); pass the flag to skip.")
+	f.Bool(flagFastSync, false, "Skip tagging and back-linking hotspots/issues with zero user changes on the source (original state, no comments, no custom tags). Defaults to false (every hotspot is tagged and back-linked). #527.")
 	f.String("default_organization", "", "SonarQube Cloud organization to migrate every project into when organizations.csv has no mapping defined. Ignored if any mapping is present.")
 	f.String("project_key_pattern", "", "Template for target project keys, built from <ORIGINAL_PROJECT_KEY> and <ORGANIZATION_KEY> (default: <ORGANIZATION_KEY>_<ORIGINAL_PROJECT_KEY>). #138")
 	f.StringSlice("exclude_branches", nil, "Glob patterns for non-main branches to skip during project data import (e.g. feature/*,bugfix/*)")
@@ -130,6 +131,7 @@ func buildMigrateConfig(cmd *cobra.Command, args []string) (migrate.MigrateConfi
 	if cmd.Flags().Changed("exclude_branches") {
 		cfg.ExcludeBranches, _ = cmd.Flags().GetStringSlice("exclude_branches")
 	}
+	applyFlagBool(cmd, flagFastSync, &cfg.FastSync)
 
 	// Default the export directory when neither config nor flag supplied
 	// one (issue #247).
--- a/go/cmd/migrate_test.go
+++ b/go/cmd/migrate_test.go
@@ -26,6 +26,7 @@ func newMigrateTestCmd() *cobra.Command {
 	f.String("target_task", "", "")
 	f.Bool("skip_profiles", false, "")
 	f.Bool("debug", false, "")
+	f.Bool(flagFastSync, false, "")
 	f.String("default_organization", "", "")
 	// Deprecated aliases — registered so tests can exercise back-compat (#406).
 	f.String("token", "", "")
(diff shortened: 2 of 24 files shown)


mannylee-sonar (approved):
Approving. The core of #527 is here and it is correct: I tested this end to end against a live SonarQube Server 2026.4.1 source and a SonarCloud target, and the eligibility classification behaved to spec in all five cases, including the new "TO_REVIEW with a user comment is still eligible" path. The speedup is real: on a project with 87 untouched TO_REVIEW hotspots, `--fast_sync` cut hotspot sync from 26.6s to 1.6s. Config-file and CLI resolution both work.

There are follow-ups I would like tracked before this ships to customers, one of which I consider important. None of them are reasons to hold the branch, but the first is a correctness issue on the default path.

### 1. The bulk index snapshot can silently drop hotspots

`buildCloudIssueIndex` takes one point-in-time snapshot per (project, branch) and every hotspot resolves against that frozen map. `PollCETask` does wait for CE, but SonarCloud's issue search is not read-your-writes consistent immediately afterwards, and the main-branch rename lands under a second before the index is built.

On `juiceshop` (87 hotspots):

| Run | Code | Tagged |
|---|---|---|
| PR, `transfer` | | **71 / 87** |
| PR, `transfer` (repeat) | | 87 / 87 |
| `main`, `transfer` | | 87 / 87 |
| PR, `sync-issues` re-run on the same project | | 87 / 87 |

One run lost 16 hotspots, and the same binary resolved all 87 on a later re-run, so it is snapshot staleness rather than a matching bug. It is intermittent, and my `main` baseline is a single run, so I am not claiming `main` is immune. But the PR is structurally more exposed: `main` issued 87 independent searches spread across the sync window, which self-heal, whereas one stale snapshot is baked into the whole project.

It is also silent. The task summary read `succeeded=71 failed=0 total=71`, and because all 87 were `excluded` the report said nothing at all.

Suggested fix: on an index miss, fall back to the pre-#527 `findCloudIssueCandidates` search for that one hotspot before recording not_found. Only misses pay a round trip, so the speed win survives on the hit path. The same fallback lets a branch whose index build failed degrade to the old path instead of being skipped wholesale, which removes the `failedBranches` silent-skip too. I have a working patch for this and can raise it as a separate PR.

### 2. `--fast_sync` gives up #423's tag, not just the back-link

`sqs-hotspot` is only ever written by `syncHotspotIssueTags` in the post-import sync, never in the fabricated scan report, so skipping that write means the tag never exists. Measured on `juiceshop`: default tagged 71/87, `--fast_sync` tagged **0/87**. The `former-hotspot` tag on those issues is SonarSource's rule tag, so it does not identify migrated findings.

#527's `[EDIT]` justifies the option on back-linking cost specifically. The tag is the cheap half, and now that resolution is a map lookup it is one write. Since #527 says untouched TO_REVIEW hotspots are the common case, a project migrated with `--fast_sync` typically ends up with zero hotspot traceability.

Worth an explicit product call rather than leaving it implicit. If tag-write volume is the real cost, `/api/issues/bulk_change` accepts up to 500 issue keys with `add_tags` and would collapse those N writes into N/500, keeping #423 and going faster than the current opt-in.

### 3. %-synced cannot reach 100% when ACKNOWLEDGED hotspots exist

`Actionable = eligible + ack`, but `a` only increments for `hotspotCategoryEligible`, so ack hotspots are in the denominator and cannot enter the numerator. My test project reported `50% of hotspots with manual changes synced (2/4)` and **Partial Migration** while every syncable hotspot synced correctly. Either count a successfully processed ack hotspot in the numerator, or render it as `2/2 synced (+2 ack inventoried)`.

Related: `result.Stats.AckDemoted = int64(counts.ack)` is now set from classification rather than outcome, so an ack hotspot that fails to resolve increments both `c` and `AckDemoted`.

### Smaller items

- **ACKNOWLEDGED no longer transitions to ACCEPTED.** `HotspotIssueStatus` maps it to `IssueStatusAccepted` and `main` applied that; this PR skips the transition, so ack hotspots land OPEN. That matches #527's wording but reverses an earlier deliberate change. Worth confirming it is intended.
- **Unresolved `excluded` hotspots are unreportable** — they return before the `b`/`c` switch, which is why the 16 lost above produced no signal.
- **`SearchAll` pagination cap.** It pages `ps=500` until `len(all) >= total`, and SonarQube rejects `p*ps > 10000`. Scope moved from (one file, one rule) to (whole project, up to 60 rules), so a large project could now hit it and fail a whole branch. Not reachable on my test data, flagging as a scaling risk only.
- **Missing #527 checklist items:** `--fast_sync` is not documented in `docs/` (`TRANSFER.md`, `MIGRATE.md`, `ADVANCED-CONFIG.md`, `HOTSPOTS-AS-ISSUES.md`), and the report has no run-metadata line recording whether it was used. Zero mentions in both generated reports.
- **`schemas/config.schema.json` lacks `fast_sync`** and sets `additionalProperties: false`, so a valid config trips schema-aware editors. Not enforced at runtime.
- **Extract cost moved rather than removed.** `enrichHotspotDetails` now hits `/api/hotspots/show` for every hotspot instead of only REVIEWED ones. Needed for the TO_REVIEW-plus-comment case and it is parallelized, but `getProjectHotspotsFull` went from 0.18s to 1.49s on `juiceshop`.
- **`hotspotResolveParams.CloudKey` is never read** (already flagged by Gitar). The fallback in item 1 would give it a use.
- **The "technical comment" filter is effectively a no-op** — every hotspot comment carries a `login`, so `hotspotHasUserComment` is really "has any comment". Fine in practice, but the comment claims more than it does.

Issue #529, opened 2026-08-19
Allow the`transfer` command to select projects by regexp


okorach-sonar:
Currently, when users want to selectively migrate specific projects, they have no other choices than running the `transfer` command for each project key that they want to transfer. This is not convenient when a fair number but not all projects needs transfer.
So:

- [ ] The `transfer` command should accept a regexp as parameter for the `--project_key` argument (or the corresponding config file parameter)
- [ ] When a regexp is passed all projects in the source SQS whose project key matches the regexp should be transferred.
- [ ] The regexp should be a full key match, it is implicitly wrapped with `^` and `$`. For instance passing `--project_key "BANKING_.+"` should only match project key that start by `BANKING_` but not projects that have the `BANKING_` string somewhere in the middle of the key
- [ ] Update documentation of the `--project_key` option to explain the change

okorach-sonar:
Fixed in #538

Issue #138, opened 2026-05-20
The project key renaming strategy should be configurable


okorach-sonar:
When migrating projects from SQS to SQC the project key should be migrated with a well defined and configurable strategy:
It should be possible, by configuration, to implement a different key renaming, namely:
- [ ] Keep the project key unchanged (to minimize impact on pipelines to redefine the project keys)
- [ ] use a unified prefix across organizations (eg the enterprise or whatever unique prefix)
- [ ] The pattern for target key creation should be buildable from 3 placeholders: <original_project_key>, <enterprise_key>, <org_key>. The project key should always be part of the target key pattern. Examples:
       - `<ORGANIZATION_KEY>_<ORIGINAL_PROJECT_KEY>`
       - `ACME_CORP_<ORIGINAL_PROJECT_KEY>`
       - `ACME_CORP_<ORGANIZATION_KEY>_<ORIGINAL_PROJECT_KEY>`
       - `<ORGANIZATION_KEY>_<ORIGINAL_PROJECT_KEY>_migrated`
       - `sqs_<ORIGINAL_PROJECT_KEY>_migrated`
- [ ] The config file should allow to specify the project key pattern in the `target` section: { "project_key_pattern": "<ORGANIZATION_KEY>_<ORIGINAL_PROJECT_KEY>" } that can use any of the 2 placeholders
- [ ] When not specified the default "project_key_pattern" should be `<ORGANIZATION_KEY>_<ORIGINAL_PROJECT_KEY>`
- [ ] It should be forbidden to use 0 placeholders in the `project_key_pattern`
- [ ] It should only be possible to use a single placeholder, if that placeholder is `<ORIGINAL_PROJECT_KEY>` and if it is combined with a prefix or postfix that is at least 5 characters: `acme_<original_project_key>` is valid, `MYCORP_<ORIGINAL_PROJECT_KEY>` is valid, `AAA_<ORIGINAL_PROJECT_KEY>` is invalid.
- [ ] If the project_key_pattern does not use the organization (eg prefix + `<ORIGINAL_PROJECT_KEY>` the tool should verify that there is no SonarQube Cloud organization that corresponds to this prefix. If there is a match, `migrate` should abort
- [ ] The `migrate` command should report in the markdown and PDF any conflict about a same project key in another organization 
- [ ] The `migrate` command should report in the markdown and PDF any key that would exceed the max project key size (400 chars to be confirmed) 
[...]

PR #540, opened 2026-08-25
Fix regtest config parsing, project scoping, and API param bugs

okorach-sonar:
## Summary
Found and fixed while running the documented regression-testing procedure against real SonarQube Server / SonarCloud infrastructure — regtest couldn't actually verify a real migration config before this.

- `regtest --config` only parsed a legacy sonarqube/sonarcloud JSON schema that nothing produces anymore. It now delegates to extract/migrate's own multi-shape config loaders and resolves the SonarCloud org from target.default_organization or organizations.csv.
- Added --project_key (mirroring transfer's #529 anchored-regex semantics) so a project-scoped transfer can actually be verified, instead of comparing the whole source server against a deliberately partial target.
- scProjectKey hardcoded the default <org>_<key> cloud key convention, ignoring target.project_key_pattern (#138). It now reads the real mapping from the latest run's createProjects output, falling back to the default convention when unavailable.
- The Issues and Hotspots checks used SQS API params that don't exist (projectKeys, projectKey) — SQS silently ignored them and returned whole-instance totals instead of per-project counts, making these checks meaningless. Fixed to the real params (components, project).
- The Global Settings check queried component=<org-key> (always 404s); org-level settings need organization=<org-key>.

## Test plan
- [x] go build ./..., go vet ./...
- [x] go test ./... (full suite green)
- [x] New unit tests for LoadConfigFile, org resolution, scProjectKey cloud-key mapping, latestRunDir, and --project_key resolution (go/internal/regtest/suite_test.go, go/cmd/regtest_test.go)
- [x] Verified end-to-end against real infrastructure: regtest --config config-ee.json --project_key "BANKING-.+" (project-scoped transfer) and a full 80-project migration, both against a live SonarQube Server + SonarCloud target


kevin-lehen-sonarsource (approved):
Apart from the 2 things that Gitar has highlighted (Cloud-key map and Hotspot checks hardcode project)
LGTM

PR #548, opened 2026-08-27
gui: honor --config for all wizard settings, fix override loss after Cancel

### LINKED ISSUE #515: `sonar-migration-tool gui` should honor the configuration file
When the GUI is started with a config file as argument, the GUI should pre-provision settings with information coming from the config file, namely:
- [ ] Source URL and token
- [ ] Source cert settings (pem file, key file, cert password)
- [ ] Export directory
- [ ] Tick box for project data migration
- [ ] Tick box for issue sync
- [ ] Target URL and Token
- [ ] Target enterprise key
- [ ] Target default organization
- [ ] Target project key(s)

- [ ] Anywhere in the GUI workflow, if a setting has not been overridden manually by the GUI user, the original config file value should persist. This is true in all scenarii (user click "Back" button, "Cancel" button etc...)
- [ ] If the user has overridden any value, this override should persist with the same pattern as described in above point (survives "back", "Cancel" and any other operation)

okorach-sonar:
## Summary
- Closes #515. `gui --config <file>` only pre-filled 5 of the 9 documented settings (URLs/tokens/enterprise key), silently dropping cert settings, default organization, and project key scoping — and lost any user-typed override (including tokens) after Cancel, since the wizard reloaded a static seed on every "Start Wizard" click instead of carrying forward what was actually submitted.
- Extends `WizardState`/the Extract and Migrate forms to cover all 9 settings (cert PEM/key/password, project key pattern with real `extract.ResolveProjectKeys` scoping, default organization), and fixes the override-loss bug by having `RunWithSeed` return its final in-memory state, which `cmd/gui.go` carries forward as the next seed.
- UI polish on top: collapsed-by-default cert settings section, required-field asterisks + hover tooltips, aligned label/field layout across every prompt, wider main pane, and a footer with the tool version on every GUI page.

## Test plan
- [x] `go build ./...` and `go vet ./...` clean
- [x] `go test ./...` — all packages pass, including new unit tests for state merge/reset, `RunWithSeed` carry-forward (incl. a direct repro of the original bug), project-key resolution, and protocol round-trips
- [x] Manual end-to-end GUI smoke test via raw WebSocket: confirmed pre-fill from config, override-survives-Cancel-within-process, and that secrets never reach `.wizard_state.json`
- [x] Visual verification via headless browser screenshots of the new field alignment, collapsed cert section, and footer


adrian-deleon-sonarsource (approved):
@adrian-deleon-sonarsource approved

PR #549, opened 2026-08-28
Stop pushing instance-scope settings at project scope, and make failures visible

joshua-quek-sonarsource:
## Why

A customer migration (org `qbe-prod`, 1142 projects) emitted **42,048 identical warnings** —
`400 Setting 'X' cannot be set on a Project` — and still exited 0 with `overall_status: success`.

Root cause: the per-project settings extract in that run held the source server's **global** settings
table, and `setProjectSettings` re-wrote it onto every project.

```
244,399 records / 1,142 projects = 214.01 each      214 × 1142 = 244,388, excess 11
```

Every project carried an identical 214-setting block, i.e. ~11 genuine per-project overrides in the
whole estate. Of the 214 per project, **49 were rejected loudly** and **~165 succeeded silently**,
becoming ~188,000 per-project overrides that shadow any later org-level change. The quiet half is the
larger problem.

The migrate side is what turned a bad input into a catastrophe: the project write path applied none of
the guards the global path has, the first 400 taught it nothing so each doomed key was retried against
all 858 projects, and `requests.log` — which seven report paths read — had no producer, so the report
showed `Failed = 0` in every section.

Full investigation: [`docs/audits/2026-08-28-project-settings-scope-leak.md`](docs/audits/2026-08-28-project-settings-scope-leak.md).

## What changed

Eight commits, each building standalone (verified per-commit in an isolated worktree).

| Commit | Change |
| --- | --- |
| `fix(extract)` | Removes a byte-identical duplicated test block that made the whole extract package fail to compile under `go test` |
| `feat(sdk)` | `IsProjectLevelRejection` (mirror of `IsOrgLevelRejection`) and `IsMangledAlreadyExists` |
| `feat(sdk)` | Per-request log transport with secret redaction; clone `http.DefaultTransport` so `MaxIdleConnsPerHost` and HTTP/2 are correct |
| `feat(migrate)` | Item-transform hook; task summaries escalate to WARN on any failure and ERROR at 100% |
| `fix(settings)` | Curated filters on the project path, runtime rejection memo, `readCustomizedSQSGlobals` hardening, leaked-globals pre-flight |
| `feat(migrate)` | `requests.log` producer, bounded phase fan-out, ALM binding pre-flight |
| `fix(gates)` | Merge records targeting one cloud gate; tolerate idempotent outcomes; log gate/op/threshold |
| `docs(audit)` | The investigation write-up |

### Two deliberate non-changes

- **`filterNonInherited` is untouched.** `ValuesAction.setInherited` is byte-identical on SonarQube
  branches 6.7→master, and on the live source (Server 2026.3.1) `?component=<project>` returns 246
  settings with **all 246** flagged `inherited: true` — the filter correctly yields zero project
  records. An earlier draft of this fix tightened it to also require `parentValue`; that would have
  **broken real overrides** (145 inherited settings carry a parent value, and 6 of 19 genuine
  overrides carry none). The customer corpus was simply not produced by this extract task.
- **The exit code and `overall_status` are unchanged.** A fully-failed task now logs at ERROR, but the
  run still exits 0. Making a failed phase fail the process is a breaking change for pipelines that
  currently pass — needs a product decision. The live run below is a good argument for making it.

Runtime classification is used rather than a pre-emptive "not in `list_definitions`" skip, deliberately:
`SetAction` **accepts undeclared keys** on a project, and hidden/secured definitions are settable but
never listed, so absence from `list_definitions` means *unknown*, not *unsettable*.

## Verification

Build, `go vet` and both test suites green; `gofmt` clean. New tests cover the rejection memo, the
curated filters, the leaked-globals detector, the gate merge, the `requests.log` format and redaction.

**Live**, source `localhost:9000` (Server 2026.3.1) → `sc-staging.io` / `open-digital-society-1`.
Driving `setProjectSettings` against real SonarCloud with a QBE-shaped corpus (13 leaked-global records):

- 7 curated keys skipped with **zero** API calls (`reason=internal-sqs-setting` / `sqs-only`)
- 5 uncurated global-only keys attempted **exactly once each**, classified by the genuine 400, then
  abandoned — one log line each
- 1 genuine setting applied → `WARN task summary … succeeded=1 failed=5 total=6`
- target project non-inherited settings 21 → 22 (only the intended probe); **zero** global-only keys
  landed. Probe reset afterwards.

Also confirmed live: `requests.log` produced (79 entries, all 5 settings-400s captured with their keys,
no secrets); `ERROR task summary task=createProjects succeeded=0 failed=1` where it was previously INFO;
ALM pre-flight fired for an unbound org before any work and stayed silent for a bound one; gates cleared
once each with **zero** 404-on-delete; no `status=0`.

**Not exercised live:** the duplicate-gate merge (this dataset has no two source gates mapping to one
cloud gate — covered by unit tests and reproduces on the committed fixture `files/2026-08-18-02/`), and
project creation, because every org on sc-staging currently refuses private projects and the tool
hardcodes `visibility: "private"`, which also blocks the already-exists path since SonarCloud validates
visibility before existence.

## Failure classification (added after the initial review pass)

The investigation kept running into the same wall: a migration says `failed=42048` and nothing tells
you whether that is the tool being wrong or SonarQube Cloud legitimately refusing something it cannot
do. Every failure now answers that.

Four classes, each carrying a plain-language `why` and a `remediation` (including "none needed" where
that is the honest answer):

| Class | Meaning |
| --- | --- |
| `by-design` | Cloud cannot do this and never will — no project scope for the setting, no Cloud equivalent for the metric. Working as intended. |
| `already-done` | The desired end state already holds. Not really a failure. |
| `environment` | Token permissions, org subscription or quota, rate limiting, connectivity, a 5xx. |
| `bug` | Rejected for a reason the tool does not recognise, so the payload it built is probably wrong. **Worth reporting.** |

An unrecognised 400 is deliberately called a **bug** rather than shrugged off — the tool sent something
Cloud would not accept and nothing else will notice. Account-state messages that would otherwise land
there (private projects not permitted, org not bound, plan limits) are matched first, so operators are
not sent chasing a defect that is really a subscription problem.

Severity now follows the **cause**, not the count: a suspected bug, or a task that achieved nothing, is
an `ERROR`; expected platform limitations stay `WARN` however many there are. The task summary reports
the breakdown, so `failed=42048, all by design` and `failed=3, all bugs` no longer look alike.

Live, verbatim:

```
level=WARN msg="setProjectSettings: setting cannot be set at project scope on SonarQube Cloud, abandoning it for the remaining projects"
  setting=sonar.lf.enableGravatar err="Setting 'sonar.lf.enableGravatar' cannot be set on a Project" status=400
  failure_class=by-design
  why="SonarQube Cloud's definition for this setting does not include the project qualifier, so it cannot be set on a project"
  remediation="none available; the setting is instance-scope-only on SonarQube Server and has no project-scope counterpart on Cloud. It is expected to be dropped"

level=WARN msg="task summary" task=setProjectSettings succeeded=1 failed=5 total=6
  failed_by_design=5 failed_already_done=0 failed_environment=0 failed_bugs=0 failed_unclassified=0

level=ERROR msg="task summary" task=createProjects succeeded=0 failed=1 total=1
  failed_environment=1 failed_bugs=0
```

Wiring detail worth reviewing: `logAPIWarn` is the single funnel all ~157 per-item failure sites already
used, so attaching the classification there lifted every site with no opt-in. The 62 sites that paired
`counter.Fail()` with `logAPIWarn(..., err, ...)` were routed through one `failAPI()` call — that pair
had already drifted in a live run, producing an item logged as `failure_class=environment` inside a
summary claiming `failed_by_design=1`. `Fail()` keeps a separate `unclassified` bucket rather than being
folded into `by-design`, because the breakdown must not claim a cause it was never told.

## Follow-ups (not in this PR)

- Decide whether a fully-failed task should fail the run.
- Report the SonarQube `String.format` double-interpolation bug upstream — it makes every
  duplicate-condition error on a percent-named metric unreadable for all users.
- Remediate the ~188,000 pinned per-project overrides already on the customer's org; `reset` only
  covers org scope today.


okorach-sonar (approved):
LGTM. Let's try to do more digestible PR in the future @joshua-quek-sonarsource .
- Open an issue describing the problem, and the desired solution
- Try to keep the PR scope small enough (this one's too large to thoroughly review)

Issue #370, opened 2026-06-09
branch last analysis date should be preserved by the migration tool


okorach-sonar:
The branch analysis date should be:
- [ ] Collected from the branch `lastAnalysisDate` on the source SQS
- [ ] Embedded in the zip that is built for the `api/ce/submit`. The way to embed the date is to be verified in the Scanner source code. The scanners accept a parameter `sonar.projectDate` that will have them built an analysis report dated from the given (older than now) date. The scanner source code is here: https://github.com/SonarSource/sonar-scanner-engine. The scanner only accept the parameter as a date, but I suppose that the analysis report can be encoded with a date and time timestamp.

SonarQube Cloud only accepts project analyzes in chronological order, so this bring extra requirements for the migration tool:
- [ ] When a project has multiple branches, the branches should be submitted in chronological order based on their `lastAnalysisDate` on the SQS side

okorach-sonar:
Fixed

PR #557, opened 2026-09-02
feat(history): migrate bounded, backdated project history (#554, PoC)

### LINKED ISSUE #554: Support migration of project history
**Disclaimer**: The would be an proof-of-concept for the moment, and it could potentially prove to not work

There could be a possibility for the migration tool to migrate project history. The high level principles of this concept would be:
- Leverage the undocumented `sonar.projectDate` scanner property that allows to back date the date of analysis
- For each project, iterate on:
   - Start from the first analysis in the project history on SQS
   - Create a synthetic analysis report with all the measures of the project (issues, files are not needed (or fake ones can be created to simply let the analysis report be accept by SQC) since they are only preserved for the last analysis of any given branch)
   - Back date the scanner report. The way to back date should be looked up in the scanner source code.
   - Iterate to next data point in the history

- In order to not migrate too many data points, and to not create too long migrations, we should limit the number of data points that we migrate and set a minimum interval between data points

[EDIT]
- All metrics that can be historized in SQC should be migrated. This includes (non exhaustive list) issue counts according to different criterias (issue counts by std experience severity (blocker issues, critical, major, minor, info), issue count by type (code smells, bugs, vulns) or sw quality (maintainability issues, reliability issues, security issues) and all other counts by type), accepted and false positive counts etc...), as well as all the other historized metrics
- It's ok if some metrics are not migrated because they are not computed in the analysis report but on the SQC side (eg probably coverage %, maybe duplication % and others). When history migration is chosen, the migration report should list in the limitations that some measures history are not migrated, citing the ones we are certain are not (coverage %, and others as we identify them)

joshua-quek-sonarsource:
## What this does

PoC implementation of project **history migration** (#554): replays a bounded set of a project's past analyses onto the target as separate, **backdated** analyses, so the target's **Activity** page shows the project's real history instead of a single "migrated today" point.

Opt-in. Off by default. With the flag unset, no extra API call is made and behaviour is byte-for-byte the pre-#554 behaviour.

Follows the approach in the issue:
- **Enumerate** history from the source via `api/project_analyses/search`, then pull each selected point's project-level measures via `api/measures/search_history`.
- **Bound** it: a minimum interval between points, and a maximum number of points.
- **Synthesize** a minimal scanner report per point — project-level measures only, no issues, plus one throwaway placeholder file component (a lone `PROJECT` component with a raw measure is rejected by the CE; the issue explicitly allows "fake ones … to let SQC accept the report").
- **Backdate** it by stamping `Metadata.AnalysisDate` with the historical date — the report-side equivalent of the scanner's `sonar.projectDate`.

## New settings

| Flag / config key | Default | Meaning |
|---|---|---|
| `--migrate_history` / `migrate_history` | `false` | Turn the feature on. Also accepted as `target.migrate_history`. |
| `--history_max_points` / `history_max_points` | `10` | Max historical points migrated per project. |
| `--history_min_interval_days` / `history_min_interval_days` | `30` | Minimum spacing between two migrated points. |

Registered on `extract`, `migrate` and `transfer`; `transfer` sets both sides from one flag.

Selection drops the newest analysis (that one is the regular current-snapshot import), applies a greedy oldest-first min-interval filter, then evenly subsamples to `max_points` so the migrated set still spans the full history range.

Source: local SonarQube Server 2026.3.1, project `okorach-oss_sonar-tools` (100 analyses, 2021-06-13 → 2026-05-09).
Target: SonarQube Cloud staging, a real organization. Run with `--migrate_history --history_max_points 8 --history_min_interval_days 30`.

**All 8 selected points were accepted by the Compute Engine (8/8 CE tasks `SUCCESS`)** and land as separate backdated analyses. `api/project_analyses/search` on the target afterwards:

```
2022-01-29T19:00:14  v1.11
2023-07-09T08:38:13  v2.7.1
2024-05-17T09:17:24  3.0
2024-12-23T11:18:32  3.8
2025-06-02T15:20:48  3.14
2025-11-14T08:43:33  3.17
2026-05-09T17:18:00  3.20
```

Dates, project versions and per-point measures match the source history. The 8th point (`2021-06-13`) was also accepted (CE task `SUCCESS`) but does **not** appear in the target's Activity — it is ~5.2 years old and SonarQube Cloud housekeeping prunes analyses past its retention window. That is target-side behaviour, not a failure of the migration; see limitations.

**No-regression check.** A separate live transfer with `migrate_history` absent was run against a different source project. The new extract task is registered but does nothing: it completes in `00:00:00.000`, issues zero API calls, and submits zero historical reports. The rest of the pipeline (project creation, current-snapshot import, main-branch rename, hotspot conversion, CE submission) proceeded exactly as before. Opting out is a true no-op.

## Bug found and fixed during live verification

The first live run failed on the very first history point:

```
CE task failed: Quality profiles with following keys don't exist in organization
[<org>]: AZ_6GBsd62_cR56qvvK5
```

The synthetic report carried a **hardcoded quality-profile key** — a literal copied from the original bring-up organization — plus a hardcoded `js` placeholder language and a hardcoded `javascript:S5542` active rule. That key exists in exactly one organization on earth, so history migration could never have worked for any other target.

This was confirmed a genuine code defect rather than an environment artifact: in the same run, against the same organization and project, the regular current-snapshot reports returned CE `SUCCESS` while only the history report was rejected.

The fix resolves the profile from the **target** organization at run time via `buildSCProfileMap` — the same lookup the regular current-snapshot import already uses — picks a placeholder language the target organization actually has a profile for (preferring common ones, deterministic fallback otherwise), and drops the fake active rule entirely (the report carries no issues, so no rule needs activating). If the organization exposes no profiles at all, history migration is skipped with an explicit warning rather than submitting a report that cannot be accepted.

Also removed in the same pass: a `NOT FOR COMMIT` live-test harness that referenced a local credentials file, and an undocumented `SMT_DEBUG_DUMP_HISTORY_ZIP` debug hook.

## Tests

- Bounding logic: empty / single-point / drops-newest / min-interval / cap-and-span / `max_points=1` / under-limit.
- Measure matching: exact timestamp, closest-prior fallback, empty metric, date-format parsing.
- Config tri-state resolution for `migrate_history`.
- Report shape, including a **regression guard asserting the qprofile key in the metadata is the target organization's** — the assertion that would have caught the bug above.
- Placeholder-language selection: prefers a common language, falls back deterministically, and reports "not possible" when the organization has no profiles.

`go build`, `go vet`, `gofmt` and the full `go test ./...` suite are clean.

## What happens to migrated points on the target

Every point the migration submits is accepted and written by the Compute Engine. SonarQube Cloud then removes some of them. On the 134-analysis test project: **133 submitted, 133 accepted, 61 present afterwards.**

That the removal is a *deletion* — rather than the CE declining old analyses, or the API not surfacing them — was established by direct observation, not inference. A throwaway project was seeded with exactly one point dated 2021-06-13 and polled sub-second:

```
15:24:14Z      CE task AaBiuH5X… → SUCCESS, analysisId AaBiuIB7_F2a6FWkA1bc
15:24:14.437Z  ┐ 12 consecutive HTTP-200 polls of
               │ /api/project_analyses/search list 2021-06-13T08:01:31, and
15:24:29.659Z  ┘ /api/qualitygates/project_status?analysisId=… returns 200
15:24:31Z      the next CE task (the regular current-snapshot import) executedAt
15:24:31.203Z  same endpoint, same query, HTTP 200 → the analysis is GONE
               and its analysisId returns 404, permanently
               (the new analysis's analysisId returns 200 — positive control)
```

228 polls on that project, zero non-200 responses. The row existed, was readable through two independent endpoints for 15.2 seconds, and was then removed.

**The control:** an identical probe whose transfer was killed 50 ms after the point landed, so no subsequent analysis ever succeeded. That 2021-06-13 analysis survived 154 seconds of continuous polling and was still alive on re-check ~18 minutes later.

So: created → readable → deleted. Not "never persisted", and not "hidden by the API".

**What is deliberately NOT claimed.** Which points get removed is not established. Two mechanisms proposed earlier in this PR were tested as classifiers and scored 63.3% and 48.6%; both were retracted rather than shipped as explanations. The oldest points do go first, consistent with the documented housekeeping window — but the test source has an 87-day gap around the apparent boundary, so the data cannot distinguish 260 weeks from 5 calendar years, and only one pre-boundary date was ever tested. Removals were also observed well inside any retention window. Treat the surviving count as something to measure per target, not to predict.

None of this is a defect in the migration: the tool's contract is that every selected point is submitted and accepted, and that holds.

## Known limitations (PoC)

- **Main branch only.** Non-main branches need the create-analysis handshake per point; generalizing that is deliberately out of scope here. History *is* currently extracted for non-main branches but not replayed — dead data worth pruning in a follow-up.
- **Measures only.** No issues, no source, no SCM, no coverage detail on historical points — by design, per the issue.
- **Not resume-safe.** History points aren't recorded in the completion map; a re-run of a branch that didn't complete can duplicate its history.
- **The target removes some migrated points.** All are accepted and written; SonarQube Cloud deletes a subset afterwards (see the section above — observed directly, mechanism not established). Nothing the migration can do about it.
- **Placeholder file component.** Each historical analysis carries one throwaway `__history_snapshot__.<ext>` component. It is never meant to be viewed, but it is present in the report.

Relates to #554.


okorach-sonar (dismissed):
I would ask for 2 changes:

1. Currently the history migration seem to only migrate the LOC measure history.
Certainly customers will look at other measures that are historized in SQS, ie any metrics that is proposed for selection in the history custom graph.

Image

Many of the same metrics can be historized in SQC

Image

2. The last analysis of each branch is not back dated.
image 
image

(Same branch last analysis is Aug 10 in SQS, and Sept 2 after migration (date of the migration))

This is something that is in the backlog of the migration tool improvement. See #370.
While we're on back dating analysis reports for history, let's back date for the last analysis too.
This back dating of last analysis should work whether the "migrate history" option is selected or not

okorach-sonar (approved):
LGTM

Issue #165, opened 2026-05-22
Some quality gates are reported as migrated multiple times in the same org


okorach-sonar:
In the migration PDF report some quality gates may be reported multiple times as migrated in the same org.
As you can see from the screenshot below, the quality gate `3 - Corp base` is reported as migrated 3 times identically in the same `fubar` organization. It should be reported only once. And the total count of migrated QG should be 2 less (because one QG migration is erroneously counted 3 times instead of 1).

<img width="1016" height="482" alt="Image" src="https://github.com/user-attachments/assets/85b37235-8b9d-4703-8823-1003c5e65842" />

PR #563, opened 2026-09-08
Report source/target rule diff for built-in QPs (#309)

### LINKED ISSUE #309: Add a comment in migration report when the source and target "Sonar Way" quality profiles are different
When migrating, the source and target "Sonar Way" QP - as well as all other built-in profiles - of a given language can be (are likely to be) different if the source and target platforms are different versions or editions. This should be reported in the actual migration report (It cannot in the predictive report as we want to be able to build a predictive report solely by looking at the source SQ data). So:

- [ ] During the `migrate` command, for the QP migration, the tool should compare each matching source and target built-in profile and report in the details column of the QP section the differences: nbr of rules added and number of rules removed (if any) between the source and target.
- [ ] When the 2 profiles are identical, the details should be empty
- [ ] When the target profile only has more rules than the source, the detail should only highlight nbr of added rules (and not say "0 removed rules")
- [ ] The QP migration status should remain "Skipped"
- [ ] The above behavior is for the actual migration report only. It does not apply to the predictive report since we don't know yet the contents of the "Sonar Way" quality profile on the SQC side when producing the predictive report

okorach-sonar:
## Summary
- Closes #309: during `migrate`, compare each source built-in quality profile against its matching target built-in profile and report the rules added/removed in the Quality Profiles section's Skipped `Detail` column (empty when identical, no "0 removed" when only additions exist, status stays Skipped). The predictive report is unaffected since the comparison needs live target-side data it doesn't have.
- Fixes three additional bugs surfaced while testing this against a real multi-org migration (`config-ee.json`):
  - **Org lookup picked the wrong/empty org**: the live-comparison org lookup kept whatever `generateOrganizationMappings` row came last for a given source server, including unbound/empty ones, silently dropping every comparison when one source server fans out to several target orgs.
  - **Duplicate "Failed" QP rows**: the `requests.log`-derived Failed bucket had no dedup and no awareness that `createProfiles` already recovers from an "already exists" 400 (the #165 fan-in case), so it rendered one duplicate Failed row per recovered attempt for entities that were, in fact, successfully migrated and already listed as Succeeded.
  - **Wrong target profile matched**: matching only checked `isBuiltIn` + `language`, so a platform shipping more than one built-in profile per language (e.g. "Sonar way" and "Sonar agentic AI" for java) paired every source built-in of that language against whichever target one happened to come first in the search results.
  - **Silent truncation past 500 rules**: SonarQube Cloud's `api/rules/search` responds with a flat top-level `total` field, not the `paging.total` the pagination helper defaults to (SonarQube Server nests it under `paging.total` too), so any built-in profile with more than 500 active rules was silently truncated to its first page.

## Test plan
- [x] `go build ./...`
- [x] `go test ./...` (full module, no regressions)
- [x] New regression tests: multi-org fan-out org lookup, requests.log Failed-vs-Succeeded dedup, multiple-built-ins-per-language matching, and target rule-fetch pagination (verified this last one fails without the fix, then passes with it)
- [x] Verified end-to-end against a real migrate run (`config-ee.json`): re-ran `--target_task compareBuiltInProfiles`, confirmed correct added/removed counts for java "Sonar way" and cross-checked against the live SonarQube Cloud API directly

### DIFF
--- a/go/internal/common/object_task_names.go
+++ b/go/internal/common/object_task_names.go
@@ -117,6 +117,7 @@ var objectCategoryTaskNames = map[string]struct {
 			"createProfiles", "setProfileParent", "restoreProfiles", "analyzeProfileRules",
 			"setDefaultProfiles", "setProfileGroupPermissions",
 			"updateRuleTags", "updateRuleDescriptions",
+			"compareBuiltInProfiles",
 		},
 	},
 	ObjectQualityGates: {
--- a/go/internal/migrate/planner.go
+++ b/go/internal/migrate/planner.go
@@ -85,6 +85,7 @@ func RegisterAll() []TaskDef {
 	all = append(all, almTasks()...)
 	all = append(all, portfolioTasks()...)
 	all = append(all, ruleTasks()...)
+	all = append(all, compareTasks()...)
 	all = append(all, deleteTasks()...)
 	all = append(all, projectDataTasks()...)
 	all = append(all, hotspotMetadataSyncTasks()...)
--- a/go/internal/migrate/tasks_compare_profiles.go
+++ b/go/internal/migrate/tasks_compare_profiles.go
@@ -1,0 +1,192 @@
+// Copyright (C) SonarSource Sàrl
+// For more information, see https://sonarsource.com/legal/
+// mailto:info AT sonarsource DOT com
+
+package migrate
+
+import (
+	"context"
+	"encoding/json"
+	"net/url"
+	"strings"
+
+	"github.com/sonar-solutions/sonar-migration-tool/internal/common"
+	"github.com/sonar-solutions/sonar-migration-tool/internal/structure"
+)
+
+// BuiltInProfileDiff is the JSONL record compareBuiltInProfiles emits for
+// each source built-in quality profile it was able to match against a
+// target built-in profile. The summary report joins these back onto the
+// existing "Built-in, not migrated" skip row (issue #309) by
+// (ServerURL, Name, Language) — the same key it already dedups extract
+// data on for the Quality Profiles section.
+type BuiltInProfileDiff struct {
+	ServerURL    string `json:"server_url"`
+	Name         string `json:"name"`
+	Language     string `json:"language"`
+	RulesAdded   int    `json:"rules_added"`
+	RulesRemoved int    `json:"rules_removed"`
+}
+
+// compareTasks returns the task that diffs source vs target built-in
+// quality profiles' active rule sets (#309).
+func compareTasks() []TaskDef {
+	return []TaskDef{
+		{
+			Name:         "compareBuiltInProfiles",
+			Dependencies: []string{"generateOrganizationMappings"},
+			Run:          runCompareBuiltInProfiles,
+		},
+	}
+}
+
+// runCompareBuiltInProfiles walks every source built-in quality profile
+// (e.g. "Sonar way") and, for each one, live-fetches the matching
+// built-in profile's active rules on the target organization to compute
+// how many rules were added/removed on SQC relative to the source. Built-
+// in profiles are never migrated (structure.addProfile drops them before
+// profiles.csv is generated), so this is the only place that ever looks
+// at their target-side rule set.
+//
+// Unlike runAnalyzeProfileRules, this task DOES call the live Cloud API —
+// it needs the target's actual rule set, which extract data can't provide.
+// That also means it only ever runs during a real migrate: the predictive
+// report's synthetic run directory never synthesizes this task's output,
+// so collectExtractSkipped simply finds no sidecar and keeps the static
+// "Built-in, not migrated" text for predictive runs.
+func runCompareBuiltInProfiles(ctx context.Context, e *Executor) error {
+	orgKeys := firstValidOrgPerServer(e)
+	activeByProfile := indexExtractByServerAndField(e, "getActiveProfileRules", "profileKey")
+
+	counter := TaskCounterFromContext(ctx)
+	return forEachExtractItem(ctx, e, "compareBuiltInProfiles", "getProfiles",
+		func(ctx context.Context, item structure.ExtractItem, w *common.ChunkWriter) error {
+			if !extractBool(item.Data, "isBuiltIn") {
+				return nil
+			}
+			name := extractField(item.Data, "name")
+			language := extractField(item.Data, "language")
+			profileKey := extractField(item.Data, "key")
+			orgKey := orgKeys[item.ServerURL]
+			if name == "" || language == "" || profileKey == "" || orgKey == "" {
+				return nil
+			}
+
+			targetProfiles, err := e.Cloud.QualityProfiles.Search(ctx, orgKey)
+			if err != nil {
+				failAPI(counter, e.Logger, "compareBuiltInProfiles: searching target profiles failed", err,
+					"organization", orgKey, "language", language)
+				return nil
+			}
+			targetKey := ""
+			for _, p := range targetProfiles {
+				// Match on name AND language, not language alone: a
+				// platform can ship more than one built-in profile per
+				// language (e.g. "Sonar way" and "Sonar agentic AI"
+				// both built-in for java). Matching by language only
+				// would pair every source built-in of that language
+				// against whichever built-in profile happened to come
+				// first in the target's search results.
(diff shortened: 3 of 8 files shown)


mannylee-sonar (approved):
# Review: PR #563 — Report source/target rule diff for built-in QPs (#309)

Approving. I checked out `fix/309-report-builtin-qp-diff` (`adcd00e`) and ran a full extract/structure/mappings/migrate cycle against a real SonarQube Server 2026.4.1 instance and a live SonarQube Cloud organization. Every claim in the PR description holds up.

## Test environment

| Side | Detail |
|:---|:---|
| Source | SonarQube Server `2026.4.1.126914`, 64 built-in quality profiles, 3 custom java profiles |
| Projects | `juiceshop`, `webgoat` (1430 source files, 2 branches) |
| Target | SonarQube Cloud, single organization (Team plan) |
| Build | `go build ./...` clean, `go vet ./...` clean, `go test ./...` 17/17 packages pass |

Portfolios were excluded via `--objects settings,permission_templates,quality_profiles,quality_gates,projects,groups` because the target organization is not part of an enterprise. Unrelated to this PR.

## Issue #309 acceptance criteria

| Criterion | Result |
|:---|:---|
| Compare each matching source/target built-in profile, report rules added/removed in the QP Details column | Pass. 61 of 61 source built-in profiles matched and diffed |
| Details empty when the two profiles are identical | Pass at the data layer (`Detail` is `""`). See note 1 below on rendering |
| Only "N added" when the target has strictly more rules, never "0 removed" | Pass. e.g. `c / Sonar way → 4 rule(s) added`, `xml / Sonar way → 15 rule(s) added` |
| Status stays "Skipped" | Pass. Every built-in row renders `Skipped` |
| Actual migration report only, not the predictive report | Pass. The predictive run directory contains no `compareBuiltInProfiles` sidecar, and the predictive PDF still reads `java / Sonar way Skipped Built-in, not migrated` with zero `rule(s)` occurrences |

### Independent verification of the diff numbers

I recomputed the java `Sonar way` diff directly against both APIs, bypassing the tool:

```
SOURCE java Sonar way rules : 605
TARGET java Sonar way rules : 629
added (target not in source) : 34
removed (source not in target): 10
```

The report says `java / Sonar way | Skipped | 34 rule(s) added, 10 rule(s) removed`. Exact match.

## The four bug fixes

All four are confirmed. For each one I reverted just that fix in a throwaway worktree and re-ran its regression test, to check the test actually guards the behaviour rather than passing incidentally.

### 1. Silent truncation past 500 rules (`TotalKey: "total"`)

Confirmed against the live APIs. `api/rules/search` genuinely differs between platforms:

```
SonarQube Cloud  top-level keys: ['p', 'ps', 'rules', 'total']            paging = None
SonarQube Server top-level keys: ['p', 'paging', 'ps', 'rules', 'total']  paging = {...}
```

With no `paging` object on Cloud, the default `paging.total` reads 0, `TotalPages` returns 0, and `GetPaginated` stops after page 1. I measured the real-world impact on java `Sonar way`:

```
target rules (with fix)  : 629  ->  added 34, removed 10   (correct)
target rules (old code)  : 500  ->  added 34, removed 139  (wrong)
```

`removed 139` instead of `10` is a badly misleading number in a customer-facing report. This fix matters.

Mutation check: dropping `TotalKey: "total"` fails `TestRunCompareBuiltInProfilesPaginatesTargetRules` with `RulesAdded: got 0 want 501`.

### 2. Wrong target profile matched (name AND language)

This reproduced naturally. The target organization ships **four** built-in java profiles:

```
Sonar agentic AI, Sonar way, Sonar way balanced, Sonar way essentials
```

The source ships two (`Sonar way`, `Sonar agentic AI`). The report correctly gives them different diffs:

```
java / Sonar agentic AI | Skipped | 2 rule(s) added, 28 rule(s) removed
java / Sonar way        | Skipped | 34 rule(s) added, 10 rule(s) removed
```

Under language-only matching both source profiles would have been paired against whichever target profile came first. Same pattern held for `ipynb`, `js`, `py`, and `ts`.

Mutation check: reverting to language-only matching fails `TestRunCompareBuiltInProfilesMultipleBuiltInsSameLanguage`.

### 3. Duplicate "Failed" QP rows

Reproduced exactly as described, including the `createProfiles` "already exists" recovery (#165 fan-in). I ran `migrate` twice against the same target so the create tasks hit 400 "already exists" and recovered, then regenerated the summary from the **same** run directory with the main-branch code and the PR code.

Old code (`main`), Quality Profiles section:

```
1 succeeded, 2 partial migration, 3 failed, 61 skipped
| MyCorp Way | ... | Failed | Quality profile already exists: java/MyCorp Way |
| Newgen     | ... | Failed | Quality profile already exists: java/Newgen     |
| Other Way  | ... | Failed | Quality profile already exists: java/Other Way  |
```

All three were also listed as Full or Partial Migration in the same table, and all three were in fact correctly migrated.

PR code, same run directory:

```
1 succeeded, 2 partial migration, 0 failed, 61 skipped
```

The effect spans every section. Total failure count across the report dropped from 11 to 4, and every dropped row was an "already exists" duplicate of a Succeeded row:

| Section | Failed (main) | Failed (PR) |
|:---|---:|---:|
| Quality Profiles | 3 | 0 |
| Permission Templates | 1 | 0 |
| Groups | 3 | 2 |
| Projects | 2 | 0 |
| Settings | 2 | 2 |
| **Total** | **11** | **4** |

The two genuine Settings failures are untouched, which is the important negative result. The guard requires both a matching Succeeded row and `"already exists"` in the error text, so ordinary failures cannot be swallowed.

Mutation check: removing the `dropFailuresAlreadySucceeded` call fails `TestCollectSummaryDropsFailuresAlreadySucceeded`.

### 4. Org lookup picked the wrong/empty org

Not reproducible in my setup, which has only one target organization. Verified by mutation instead: swapping `firstValidOrgPerServer` back to `buildServerOrgLookup` fails `TestRunCompareBuiltInProfilesMultiOrgFanOut` with `expected exactly 1 diff record despite multi-org fan-out, got 0`, which is precisely the "silently dropping every comparison" symptom described. The reasoning in the doc comment is sound: built-in rule content is identical across organizations on the same target platform, so any one bound organization is representative.

## Notes, none blocking

1. **Identical profiles render as `Built-in`, not empty.** `formatBuiltInProfileDiff(0, 0)` correctly returns `""`, but `skippedDetails` (pre-existing behaviour in `pdf.go`) falls back to the skip-reason label when `Detail` is empty. The rendered cell reads `Built-in`. I think that is nicer than a blank cell, so I would not change it. Two side effects worth knowing:
   - A profile that is identical and a profile that could not be matched to any target profile both render `Built-in`, so they are indistinguishable in the report. In my run all 61 matched, so this did not bite, but a customer on an older target platform could see it.
   - The predictive report says `Built-in, not migrated` while the real migrate report says `Built-in` for the same identical profile. Minor wording drift between the two reports.

2. **Residual false-positive Failed rows on a re-run.** `migration-scanners` and `migration-viewers` still appear as Failed with "already exists" on a second `migrate`. They are created by `createMigrationGroups`, so they never land in the Succeeded bucket, and `dropFailuresAlreadySucceeded` has nothing to match them against. Strictly better than before (3 group failures down to 2), but the same class of noise remains for the tool's own internal groups. Worth a follow-up issue rather than expanding this PR.

3. `compareBuiltInProfiles` cost 3.9s for 61 profiles on this instance, one `Search` per organization plus one paginated `api/rules/search` per matched profile. Fine at this size. On a very wide instance migrating into many organizations it is one extra rules-search round trip per built-in profile per organization, which is worth keeping an eye on but not a concern here.

## Test steps run

```bash
git checkout fix/309-report-builtin-qp-diff
make build && (cd go && go vet ./... && go test ./... -count=1)

sonar-migration-tool extract  --config config.json --project_key "juiceshop|webgoat"
sonar-migration-tool structure --config config.json
sonar-migration-tool mappings  --config config.json
sonar-migration-tool migrate   --config config.json \
  --default_organization <org> --enterprise_key <ent> \
  --objects settings,permission_templates,quality_profiles,quality_gates,projects,groups
sonar-migration-tool predictive-report --config config.json
```

Then a second `migrate` against the same target to trigger the "already exists" recovery paths, and a report regenerated from that one run directory with both `main` and this branch for a like-for-like diff.

The sandbox organization was cleaned between runs and left empty afterwards.

Issue #564, opened 2026-09-09
Improve ETA estimation


okorach-sonar:
The current algorithm for ETA estimation is very rough. This should be improved to give a better estimation of when a migration will complete.
So:
- [ ] Look at the logs of a migration (`extract` and `migrate`)
- [ ] From these logs, infer the real duration of each operation (rather than % roughly defined with the previous algorithm)
- [ ] Revisit ETA calculation based on the real duration of these operations
- [ ] Also take into account the time taken for project history migration when this option is chosen

PR #572, opened 2026-09-16
ETA estimation improvements (#564): per-call duration logging + seeded task durations

okorach-sonar:
## Summary
- Clocks each API round trip in `debugTransport.RoundTrip` (SDK) and threads the duration through `DebugLogFunc` — `--debug` logs now include a `duration_ms` field on every `http request` / `http request failed` entry, alongside method, URL, headers, and status
- Seeds `Tracker` with real, per-task expected durations mined from actual extract/migrate run logs (`SeedTaskDurations` / `ExpectedTaskDuration`) instead of treating every task in a category as equally costly, with a `DefaultTaskDuration` fallback for small orchestration tasks
- Gives project-history replay (#554) its own tracked unit of work in the overall ETA (`migrateProjectHistory` pseudo-task, incremented per historical point submitted) since it previously ran invisibly inside `importProjectData`
- `migrate.sh`: passes through extra CLI args and supports toggling history migration, for local testing of this feature

## Test plan
- [x] `go build ./...` (both `go/` and `lib/sq-api-go/`)
- [x] `go vet ./...`
- [x] Full `go test ./...` for both modules — all passing
- [x] Manually verified `duration_ms` appears on both success and failure debug log entries


kevin-lehen-sonarsource (approved):
LGTM
Feels like the migrate.sh should have some guardrails to check if parameters have something filed in.
e.g. -c is expecting something but if nothing is provided it could break the script

PR #575, opened 2026-09-16
Live end-to-end "user simulated" smoke suite for every CLI command

joshua-quek-sonarsource:
**What changed:**
Short version - Live Smoke E2E test suite has been built.

**Outcome:**
- ✅ Safer releases — every command gets tested against the real thing before we ship
- ✅ Locked down — 5 separate safeguards so it can't accidentally touch production
- ⚠️ Not fully proven yet — the "delete stuff" safety check has never actually been run, and we've never watched it catch a mistake on purpose

**Bottom line:** Big upgrade to test quality.


joshua-quek-sonarsource:
Tracking issue for this suite and its remaining work: #577 — including the two items I'd want closed before this merges (running Tier 2 against staging at least once, and confirming the report-accuracy assertions actually go red when `4a76134` is reverted).

joshua-quek-sonarsource:
Ok I am getting claude to fix all of the gitar bugs now first.
Then rebase before merge (later).

joshua-quek-sonarsource:
**Summary of latest commits (plain English):**

- **New "yellow" result for the regression tester.** Some differences between SonarQube Server and SonarQube Cloud aren't bugs — they're just how the two products work (e.g. Cloud auto-names its default quality profile differently, has its own rule versions, and this tool always makes migrated projects private). We now label these clearly as `SQS_AND_SQC_FEATURE_DIVERGENCE` and show them as a yellow warning instead of a red failure, so real bugs never get lost in the noise — but a real bug always still shows red, even alongside yellow ones.
- **Tidied up docs**: moved the 25 product spec files from `roadmap/specs/` to a simpler top-level `specs/` folder (no content changed).
- **Fixed a few test bugs**: our automated smoke tests now actually exercise real project migrations (not skip them), catch it if a CLI command is silently added/removed, and fixed a text-cutoff bug in one of our reports.

No behavior change for end users migrating a real project — this is about making test results clearer and more trustworthy.

okorach-sonar (approved):
LGTM

PR #578, opened 2026-09-16
Reduce API call pressure on SonarQube Cloud

### LINKED ISSUE #571: Optimize (reduce) API calls to lower the pressure of SonarQube Cloud and avoid rate limiting
SonarQube Cloud has a rate limiting mechanism that triggers when a particular IP is hitting the APIs to hard.
Also, even when API calls are accepted, this puts some pressure on the SonarQube Cloud operations.
So we want to make sure that we reduce the number of API calls to the strict necessary. this means:
- [ ] Reduce polling frequency of calls to `api/ce/task`, wait twice longer between polls
- [ ] Cap the number of comments applied to an issue to the last (most recent) 5 comments. Add a tool configuration parameter `--max_issue_comments` to override the value 5 with other value (less or more). Don't allow any value higher than 20
- [ ] Make sure that all issue tags are applied in a single `api/issues/set_tags` API calls (which does not seem to be the case right now)

okorach-sonar:
## Summary
Closes #571. Three changes to cut back on API calls a migration run makes against SonarQube Cloud.

This PR's implementation was consolidated with the independent, parallel implementation from PR #579 (`fix/571-reduce-api-calls`) into a single history — the two branches solved the same issue concurrently, so this merges #579's code in, keeping one canonical implementation here rather than shipping duplicate work.

- **`api/ce/task` polling doubled** (5s -> 10s) via `scanreport.CEPollInterval`, a package var so tests can drive it to near-zero instead of paying the real delay.
- **Issue/hotspot comments capped**: new `--max_issue_comments` flag on `migrate`, `transfer`, and `sync-issues` (plus the matching `max_issue_comments` config-file field across all 4 config shapes). Default `5`, max `20` (rejected above that via `ValidateMaxIssueComments`), and `-1` is a documented sentinel to disable the cap entirely. Only each issue/hotspot's most recent N comments (by `createdAt`) are migrated via `api/issues/add_comment`, via a shared `capIssueComments` helper used by both the issue-sync and hotspot-as-issue-sync paths.
- **`api/issues/set_tags`**: verified it already sends every tag in a single call per issue (comma-joined, not a per-tag loop) — confirmed by `TestSyncIssueTagsSendsUnion`. No code change needed for this checklist item.

Also updated `WHATSNEW.md`, `docs/ADVANCED-CONFIG.md`, `docs/SYNC-ISSUES.md`, and `schemas/config.schema.json` for the new flag.

## Test plan
- [x] `go build ./...` (both `go/` and `lib/sq-api-go/`)
- [x] `go vet ./...`
- [x] Full `go test ./...` for both modules — all passing, `migrate` package suite stays fast (~16s) thanks to test-only overrides of `CEPollInterval`
- [x] Comment-capping logic (`capIssueComments`, `syncIssueComments`), config-file resolution for `max_issue_comments` across all shapes, `applyDefaults` defaulting/sentinel handling, and rejection of values above 20 in `RunMigrate`/`RunSyncIssues`
- [x] Manually verified `--help` output and the rejection path on all three commands (`migrate`, `transfer`, `sync-issues`) via the built binary


okorach-sonar:
gitar auto-apply:on

mannylee-sonar (approved):
## What this PR set out to fix

This PR cuts down how many API calls the migration tool makes against SonarQube Cloud. It does three things:
- Waits 10 seconds between task-status checks instead of 5 seconds.
- Only copies the 5 newest comments per issue (adjustable, max 20), instead of every comment ever written.
- Confirms tags were already sent in one batch call (no code change needed there).

## What I tested

I built this PR's code and ran it for real: copied the `juice-shop` test project from a source server to a sandbox SonarQube Cloud org.
- Full build: clean. All existing automated tests pass, in both modules.
- Ran a full copy of the project, start to finish.
- Added fake comments to test issues (7 on one, 6 on another, 8 on a third) to check the new comment limit for real.
- Tried the comment limit at its default, and set to a custom value of 3, and tried an invalid value (21) to check it gets rejected.

## Per-fix results

| Claimed fix | Verified? | Evidence |
|---|---|---|
| Wait 10s between task checks (was 5s) | YES | Code sets the wait to 10 seconds. Automated tests check the wait logic. Live run finished with no errors. |
| Copy only the newest N comments per issue (default 5, max 20) | YES | Issue with 7 comments → kept the newest 5. Issue with 6 comments → kept the newest 5. Issue with 8 comments, limit set to 3 → kept the newest 3. Right ones kept, in the right order, every time. |
| Reject a comment limit above 20 | YES | Setting the limit to 21 was refused immediately, with a clear error message. |
| Tags already sent in one call (no code change claimed) | YES | Confirmed by an existing automated test that already checks this. |

## Problems surfaced

Two things came up during testing. Neither one is caused by this PR.
- **Running the full copy twice in a row on an unchanged project fails.** SonarQube Cloud says the second upload looks "older" than the first and refuses it. Repro: run `transfer` on the same project twice back-to-back with nothing changed on the source. This is in a part of the code this PR does not touch, so it's a pre-existing issue for the team to track separately, not a reason to block this PR.
- **Once an issue is fully copied once, the comment limit is locked in for that issue.** Re-running later with a bigger limit won't add the older comments back. Repro: sync an issue with the default limit, then re-run with a bigger limit. This is old, on-purpose behavior (skip issues already marked done), not new in this PR. Worth knowing, not a blocker.

## Recommendation: Approve

The PR does what it says it does. I proved this by running it for real, not just by reading the code. The two problems found above already existed before this PR and are not things it broke.

PR #579, opened 2026-09-16
Reduce SonarQube Cloud API pressure (#571)

### LINKED ISSUE #571: Optimize (reduce) API calls to lower the pressure of SonarQube Cloud and avoid rate limiting

okorach-sonar:
## Summary
Closes #571 — reduces the number/frequency of API calls sonar-migration-tool makes against SonarQube Cloud during migrate/transfer/sync-issues, to lower rate-limiting risk and operational pressure.

- **Slower CE polling**: `PollCETask`'s `api/ce/task` poll interval is doubled from 5s to 10s (`internal/scanreport/submit.go`), via a package-level `ceTaskPollInterval` variable so tests can override it and stay fast.
- **Capped issue/hotspot comments**: new `--max_issue_comments` flag (default 5, max 20 — rejected above that) caps how many of the most-recent source comments are replayed per Cloud issue/hotspot, instead of the whole thread. Wired through `migrate`, `transfer`, and `sync-issues` commands and all 4 documented config-file shapes. Capping is done by a new `capIssueComments` helper (sorts by `CreatedAt`, keeps the tail) used by `syncIssueComments`, shared by both the issue-sync and hotspot-as-issue-sync paths.
- **Single `set_tags` call verified**: audited `syncIssueTags` and `syncHotspotIssueTags` — both already merge tags and call `/api/issues/set_tags` exactly once per issue/hotspot. No code change was needed for this checklist item.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` clean on all touched files
- [x] `go test ./...` — full suite passes
- [x] New unit tests: `TestCapIssueComments`, `TestSyncIssueCommentsCapsToMostRecent`, `TestSyncIssueCommentsUncappedReplaysEveryComment`, `TestValidateMaxIssueComments`, `TestMigrateConfig_ApplyDefaultsFillsMaxIssueComments`, `TestLoadMigrateConfigFile_MaxIssueCommentsAllShapes` (all 4 config shapes)
- [x] Manually verified `--help` output and validation error for `migrate`, `transfer`, `sync-issues`
- [x] Confirmed `TestPollCETaskSuccess`/`TestPollCETaskFailure` still run fast (<1s total) despite the doubled interval


okorach-sonar:
gitar auto-apply:on

okorach-sonar:
Duplicate of #578

Issue #400, opened 2026-06-16
Migration from 9.9 does not migrate all issues


okorach-sonar:
Making a test migration from SonarQube Server 9.9 to SQC I noticed that the issues are not all migrated. 
On the source SQS 9.9: 188 open issues, 2 won't fix and 3 FP
On the target SQC: 41 open issues, 1 won't fix and 1 FP

The problem does not seem to be on the issue status count, simply that not all issues are migrated (42 issues migrated our of 191)

See screenshot below:

<img width="1691" height="962" alt="Image" src="https://github.com/user-attachments/assets/adfbd8e4-fd2f-4d5e-9f63-6642c4aea97b" />

PR #580, opened 2026-09-17
feat(extract): slice issues by creation date past the 10,000-result cap (#574)

### LINKED ISSUE #574: Issue extraction silently truncates at SonarQube's 10,000-result cap (no slicing, unlike CloudVoyager)
## Problem

SonarQube Server/Cloud's issue search API has a hard ceiling of **10,000 results** (an Elasticsearch limit on the search backend). Today, `sonar-migration-tool` treats this as a silent stopping point rather than working around it:

```go
PageLimit: 20, // SonarQube caps at 10,000 results
```

This appears in `go/internal/extract/tasks_projectdata.go` (around lines 113 and 315) and `go/internal/extract/tasks_history.go` (around line 154) — `PageLimit: 20` × `ps: 500` = 10,000. Once a project's true issue count exceeds that, extraction stops there with **no warning, no chunking, and no indication to the user that data was truncated**.

There is currently no date-range, rule-based, or facet-based partitioning logic anywhere in the extractor to get past this ceiling.

## Impact

Any project whose actual issue count exceeds 10,000 will get a **silently incomplete migration** — up to 10,000 issues extracted/migrated, with the remainder simply dropped. For large, long-lived projects this is a real correctness risk for the whole migration.

## Prior art: CloudVoyager already solved this

Per this project's own `CLAUDE.md` directive to harvest features from CloudVoyager (the predecessor project), this exact problem was already solved there:

**[cloudvoyager PR #112 — "Issue #94 - fetch 10k issues limit sqc"](https://github.com/sonar-solutions/cloudvoyager/pull/112)** (merged 2026-04-18)

- Root cause: SonarCloud's `searchIssues` helper used naive page-loop pagination that silently truncates at 10K.
- Fix: a shared `fetchWithSlicing` helper that probes the total result count for a query, and — when it exceeds 10K — **recursively splits the query into narrower creation-date windows**, deduplicating issues at overlapping window boundaries, until every slice is under the cap.
- Wired into all four pipeline variants (`sq-9.9`, `sq-10.0`, `sq-10.4`, `sq-2025`) via each `search-issues.js`, with the shared logic living under `src/shared/utils/search-slicer/...`.

## Proposed fix

Port the recursive creation-date-window slicing approach from CloudVoyager PR #112 into the Go extractor:

- Detect when a query's total result count would exceed (or is at) the 10,000 cap.
- Recursively bisect the query by `createdAfter`/`createdBefore` date windows until each slice's count is safely under the cap.
- Dedupe issues at window boundaries.
- Apply this to both `go/internal/extract/tasks_projectdata.go` (issue/hotspot extraction) and `go/internal/extract/tasks_history.go`.
- Surface a clear log/warning if a project is large enough to require slicing, so this is visible rather than silent.

Per `CLAUDE.md`'s smoke-suite mandate, this would also need coverage added to `go/smoke/` (a case exercising a project whose issue count exceeds a single page/window, or a unit-level test of the slicing logic itself, plus any new report fields if truncation status is surfaced in the report).

## Search notes

No existing issue or PR in this repo currently tracks this (checked issues/PRs #348, #350, #356, #371, #400/#403, #541/#544, #554/#557 — all unrelated: sync performance, OOM, filter-param bugs, or history bounding, not the issue-search 10k cap).

mannylee-sonar:
## The problem

SonarQube's search API never returns more than 10,000 results for one query. To stay under that limit, the issues extract stopped after 20 pages of 500, and said nothing about stopping.

So any project with more than 10,000 issues got migrated with only 10,000 of them. The rest were dropped. The migration report then printed "10,000" as if that were the whole project.

Measured on a real instance before this change:

- `opencv` extracted 10,000 of 73,293 issues. 63,293 were dropped, which is 86% of the project.
- `zvec` extracted 10,000 of 14,903. 4,903 dropped.

No warning, no log line, nothing in the report. The tool exited 0.

Closes #574.

## The change

**1. Truncation is now reported.** The server already tells us the real total on the first page, and we were reading it and throwing it away. Now, whenever we fetch fewer results than the server says exist, we say so: on the console during extract, in a new `extract_truncation.json` file, and as a Limitations bullet in the migration report. A run that lost nothing writes nothing.

**2. Large issue sets are fetched in date ranges.** If a project has more than 10,000 issues, we split the query by issue creation date and keep splitting each range in half until every range is small enough to fetch completely. Then we combine the results and drop any duplicates.

Projects under 10,000 issues are completely unaffected. Same number of requests as before, and no date filter is ever added.

**3. Issues only, not hotspots.** The hotspots API has no date filter at all. Worse, if you send one anyway it returns HTTP 200 and quietly ignores it, so every split would come back with the same number and the code would split forever while fetching the same truncated data. Three separate guards make that impossible to do by accident.

## Verified against a real SonarQube, version 2026.4.1

| project | extracted | reported as lost | total | what the server says exists |
|---|---|---|---|---|
| curl | 3,009 | 0 | 3,009 | 3,009 |
| zvec | 14,662 | 0 | 14,662 | 14,662 |
| drawio | 18,243 | 7,919 | 26,162 | 26,162 |
| opencv | 73,293 | 0 | 73,293 | 73,293 |

Every row adds up to the server's own count. No duplicates in any run.

I also logged every HTTP request through a proxy and checked them: no empty date range, no date without a time, no date parameter ever sent to the hotspots API, and the project filter stayed `componentKeys` throughout (see #400). Re-running with `--extract_id` keeps the truncation record instead of overwriting it.

## What this still cannot fix

A date range cannot be narrower than one second. If more than 10,000 issues share the exact same creation second, we cannot split them apart.

That is not rare. When a project is first analysed, every existing issue gets stamped with that one moment. `drawio` has 17,919 issues on a single second.

For that project we now extract 18,243 issues and report exactly 7,919 as unreachable, naming the second they are in. The loss is measured and visible. It is not recovered.

A follow-up will split such a second by issue type and severity instead of by time. That is not in this PR because no project we have access to has 10,000 issues in one second, so that code could only be tested against a fake server. Everything in this PR was tested against a real one.

## Worth a close look in review

**How the date ranges are cut.** Each range includes its start time and excludes its end time. When a range is split in two, both halves use the exact same split point, so nothing can fall in the gap and nothing gets counted twice.

This is worth checking because CloudVoyager gets it wrong. It adds one millisecond to the start of every range, which someone added believing the ranges overlapped. They never did. And when that millisecond rolls over into the next second, every issue created during that second is dropped and nobody finds out. We do not copy that, and there is a test to keep it that way.

**The first and last ranges have no boundary at all.** The very first query sends no date filter, so it is identical to what the tool sends today. Splitting only ever adds boundaries in the middle, never at the two ends.

That matters because it removes a whole class of bugs. There is no guessed start date, no dependence on the clock of the machine running the tool, and no gap for issues created while a long extract is still running.

**`PageLimit: 20` is still there on purpose.** 20 pages of 500 is exactly the largest request the server will accept. Removing it does not fetch more, it makes the server return HTTP 400 and kills the whole task.

**Two bugs were found by reviewing this diff, and both are fixed here.** Worth knowing what they were:

The first one lost data. To pick split points we ask the server for the oldest and newest issue dates. We then used the newest issue's timestamp as the end of the range. But the end of a range is excluded, and that timestamp belongs to a real issue, so the last range ended up covering two seconds instead of one while still being treated as too small to split. Issues that could have been fetched were dropped, and the report blamed the API limit for it.

The second one made things worse than before. If a server rejects date filters, the tool used to return 10,000 issues and finish. With the first version of this change it returned zero and aborted the entire run, including every other project. It now falls back to the old behaviour.

Both have tests.

## Deliberately not included

- Splitting hotspots, the component tree, or analysis history. None of them have a usable date filter. They report truncation instead.
- Catching the HTTP 400 at the 18 other paginated calls that have no page cap. Doing that would turn a loud failure into a silent one in `ListAllProjectKeys`, which decides which projects get migrated at all.
- Changing the per-branch "Issues" number in the report. That number is just a count of what landed on disk, and changing it is a bigger change to how the report decides outcomes.
- A `--fail_on_truncation` flag. In `transfer`, extract and migrate run in one process, so exiting early would skip the report and hide the loss more effectively than the current silence does.


okorach-sonar (commented):
@mannylee-sonar 
This looks like a regression because this was supposed to be handled.
When the SQC search hits the 10000 issues limit, the tool is supposed to slice per date range, and slice further by dichotomy, down to a range of 1 day
about

> What this still cannot fix
> A date range cannot be narrower than one second. If more than 10,000 issues share the exact same creation second, we cannot split them apart.

I am surprised by this statement:
1) To my knowledge you can only search by date range, not datetime, so the best granularity is 1 day, not 1 second
2) There is a way to split further by adding other filters on the search: The 1st additional filter we can leverage is the ruleId, but we can also add if needed in 2nd the directory then the file, 3rd the status (FP and Accepteds)

All the above was tested at some point, I will double check that again. I may attach sample projects that precisely reproduce those corner cases
- more than 10k issues
- more than 10k issues of a same rule
- more than 10k issues of a same rule in a same directory
- more than 10k issues of a same rule in a same file

See attached zip files. You can scan those projects in SQS and try to migrate them.
It's important during SQS scan to not have SCM data so that all issues are dated from the same day (date of analysis).

[12k-issues-structured.zip](https://github.com/user-attachments/files/32335533/12k-issues-structured.zip)

[12k-issues-flat.zip](https://github.com/user-attachments/files/32335537/12k-issues-flat.zip)
---
EDIT: Actually the above slicing behavior was implemented in cloud voyager but not in `sonar-migration-tool` 
So this feature is much needed.
It would be important to add the slicing on ruleid, directory/file and status in further PRs

okorach-sonar (approved):
LGTM

mannylee-sonar:
@okorach-sonar Thanks for the reproducers, they were really useful. I scanned both in SQS, ran the extract, and then ran a full migration to a staging SQC org. Sharing what I found, including a couple of things that surprised me.

### Both projects migrate end to end

Extract, three separate runs, identical each time:

```
12k-issues-flat:       total=12000 collected=12000 windows=2 requests=33 expectedLoss=0 drift=0
12k-issues-structured: total=12000 collected=12000 windows=2 requests=33 expectedLoss=0 drift=0
```

12,000 unique issue keys on disk per project, no truncation record. Both packaged at `issues=12000`, both landed on SQC as **Full Migration**, and I confirmed `violations = 12000` on each target project. So these two cases come through complete.

### The granularity is actually seconds, not days

This was the thing that changed the picture for me. `createdAfter` / `createdBefore` do take a full datetime, so the floor is one second. The format is fussy: it rejects `Z`, rejects milliseconds, and rejects `+00:00`, and wants `2006-01-02T15:04:05-0700`. Measured on your projects:

| bound | result |
|---|---|
| date only, `2026-09-18` to `2026-09-19` | 12,000 |
| `...T09:10:00+0000` to `...T09:10:01+0000` | 6,000 |
| `...T09:10:02+0000` to `...T09:10:03+0000` | 6,000 |

Even with no SCM data the scan spans two seconds, because the scan itself takes longer than a second. At day granularity your projects would be one unsplittable 12,000 bucket, exactly the failure you were expecting. At second granularity they split into two 6,000 windows, which is why the extract completes.

### One thing worth flagging on the SQC side

Checking the target turned up something I did not expect. On SQC, `/api/issues/search` reports `paging.total = 10000` for both migrated projects, but the computed measure reports `violations = 12000`, and slicing the same query by date on SQC sums to 12,000. SQS behaves differently: it reported the true 73,293 for opencv.

So SQC appears to clamp the total at exactly 10,000 rather than reporting it. That is worth knowing, because our split predicate is `total > 10000`, and a clamped 10,000 does not satisfy it. Extract reads SQS so it is not affected today, but anything reading issues back from SQC cannot distinguish 10,000 from 12,000 by the total alone. Happy to open a separate issue for that if it is not already known.

### On ruleId, one wrinkle worth knowing

I like the idea of secondary axes, but measuring them turned up a catch. `12k-issues-structured` happens to have all 12,000 issues under a single rule (`python:S1481`), so a rule split leaves one oversized bucket, and a rule cannot be bisected the way a date range can. Real projects hit this too: drawio has 15,874 issues under `javascript:S3504`, and paging that single-rule query past 10,000 still returns HTTP 400.

The facet cap is the other constraint. Rule and file facets both stop at 100 values, so on opencv the rules facet covered 97.3% of 73,293 issues, leaving 1,984 invisible. Pulling the full list from `/api/rules/search` avoids that, but there are 8,353 rules on my instance, so it trades 33 requests for a few thousand.

Comparing the axes across your two projects:

| axis | 12k-issues-flat | 12k-issues-structured |
|---|---|---|
| creation date | 2 windows, 6,000 each ✅ | 2 windows, 6,000 each ✅ |
| ruleId | 2 rules, 6,000 each ✅ | 1 rule, 12,000 ❌ |
| directory | 1 directory, 12,000 ❌ | 2 directories, 6,000 each ✅ |
| file | 120 files, 100 each ✅ | 120 files, 100 each ✅ |

That is roughly why we landed on date first: time is dense, so the recursion always has somewhere to go. Rule and directory are coarse, so a single oversized bucket has no fallback.

### On CloudVoyager

Confirming your edit. I read `src/shared/utils/search-slicer/` in cloudvoyager#112 and it is the same recursive creation-date approach, so we ported it rather than inventing something new. A few things changed on the way in: open outer bounds instead of a hardcoded 2006 epoch, a shared window boundary to avoid a millisecond rounding gap at the seams, streaming to disk instead of accumulating in memory, and a guard so this never runs against `/api/hotspots/search`, which returns 200 with an unchanged total for unknown params.

### Where I think you are right

The same-second case is real, it just is not what these two projects hit. If a scan stamps more than 10,000 issues in one second, bisection runs out and we report the exact shortfall rather than silently dropping.

For that, **file looks like the stronger next axis than ruleId**. It is the only one in the table above that covers both of your projects, files are enumerable from the component tree we already fetch so the 100-value facet cap does not apply, and file size is bounded in a way rule count is not. SPEC-006 already flags the component axis as the better long term option. Happy to open a follow-up and pick this up if that sounds right to you.

PR #585, opened 2026-09-17
feat(migrate): dynamically adjust concurrency for SonarQube Cloud rate limits (#573)

### LINKED ISSUE #573: Adjust concurrency dynamically to respect SonarQube Cloud rate limiting but also optimize migration speed
This issue only relate to the migrate command when the target is SonarQube Cloud (and eventually to extract when we will allow SQC to SQC migration).

SonarQube Cloud API rate limiting is 8000 API calls/5 minutes/IP
Any attempt above this is rejected with an HTTP code 429.
So rather than pushing the API calls to the limit, it makes more sense to throttle to a pace that will avoid rate limiting.
So:
- [ ] Spread API calls to avoid exceeding 8000 calls/5 min. Limit to 1500 calls/min, as a sliding window. If that limit is reached, wait until the sliding window frees up API calls, then resume
- [ ] Dynamically adjust concurrency to stay close to 1500 calls/min but not exceed that. If the SQC API calls take longer time to execute, increase the number of concurrent API calls. If the API calls are fast, reduce concurrency.
   -  Example 1. Situation where SQC API calls take, in average 200ms. A concurrency of 10 would generate 60 * 10 / 0.2 = 3000 API calls / min. This is too much, concurrency should be reduce to 5, so that the nbr of API calls in the minute is 60 * 5 / 0.2 = 1500 API calls / min
   - Example 2. Situation where SQC API calls take, in average 2s / call. A concurrency of 10 would generate 60 * 10 / 2 = 300 API calls / min. This is too little, the migration will be slower than what is possible, concurrency should be increased to 50, so that the nbr of API calls in the minute is 60 * 50 / 2 = 1500 API calls / min
- [ ] The concurrency should be reevaluated on a very regular basis (every 30 secs) to adjust to shorter or longer API calls durations. When re-evaluate an INFO log should state what is the new API calls concurrency value. 
- [ ] Deprecate the `--concurrency` API for all SonarQube Cloud API calls. When this parameter is used against an SQC (`migrate` or `transfer` command), log a deprecation warning
- [ ] Add a `--api_max_rate_per_min` parameter, to throttle API calls. By default this parameter should be set to `1500`. It can be set to any value between `100` and `1500` (any value outside of this range should trigger a warning and abort the migration tool command)

okorach-sonar:
## Summary

Closes #573.

SonarQube Cloud rate-limits at 8000 API calls/5min/IP (HTTP 429 above that). This adds proactive throttling and self-tuning concurrency for `migrate`/`transfer`/`sync-issues`/`reset` (the SonarQube Cloud-facing commands) instead of relying only on a fixed `--concurrency` worker count and reactive 429 backoff:

- **Sliding-window rate limiter** (`lib/sq-api-go/ratelimiter.go`): a new `throttleTransport` inserted innermost in the HTTP transport chain — wrapping only the base `http.Transport` — so every physical attempt (including retries) is throttled and timed against a configurable calls/min cap.
- **Dynamic concurrency** (`go/internal/migrate/concurrency.go`): `ConcurrencyLimiter` replaces the old `Executor.Sem` capacity-carrier channel. In dynamic mode it recalculates every 30s from a tumbling latency average, targeting the configured rate (`ceil(rate × avgLatency / 60)`, clamped to `[1, 100]`), and logs a DEBUG line on every recalculation. An explicit `--concurrency` still works as a fixed override.
- **New `--api_max_rate_per_min` flag** on `migrate`, `transfer`, `sync-issues`, `reset` — default 1500, valid range `[100, 1500]`, hard-aborts (not clamps) out of range.
- **`--concurrency` deprecation**: logs a warning when explicitly passed against these Cloud targets, but keeps working as a fixed value.
- **Smoke suite**: added `--api_max_rate_per_min` to the tier0 CLI-contract flag table and a `TestTier0_ExitCodeMatrix` case for the out-of-range abort, verified against the real binary.

## Test plan

- [x] `go build ./...` / `go vet ./...` / `go test ./...` pass with zero regressions in both `go/` and `lib/sq-api-go/` modules
- [x] New unit tests: sliding-window limiter (concurrency + context cancellation + httptest-backed latency sampling), concurrency formula/clamp table tests, `ConcurrencyLimiter` dynamic convergence + DEBUG log emission + fixed-mode no-op, CLI `validateAPIMaxRatePerMin` table test, CLI deprecation-warning test
- [x] `go test -tags smoke -run TestTier0` passes, including the new exit-code case
- [ ] Live SonarQube Cloud run to confirm rate limiting holds under real load (not exercised in this environment — no SonarQube Cloud connection available here)


okorach-sonar:
gitar auto-apply:on

mannylee-sonar:
I fixed the two blockers on this PR. I'm not approving or merging. That's for someone else.

**1. Merge conflicts**

I merged main into this branch. That cleared the conflicts in:
- go/cmd/sync_issues.go
- go/cmd/transfer.go
- go/smoke/tier0_cli_test.go
- lib/sq-api-go/options.go

These weren't real conflicts. This branch already had main's #587 change (the --insecure flag), just under a different commit ID. Git saw two different IDs and called it a conflict, even though the code was already the same.

I kept this branch's version in all four files. The result is byte-for-byte the same as before the merge. No behavior changed.

**2. Cognitive Complexity issue**

SonarQube Cloud flagged one function for being too complex: [propagateGlobalsToProjects, complexity 18, limit 15](https://sonarcloud.io/project/issues?id=SonarSource_sonar-migration-tool&pullRequest=585&s=IMPACT_RANK&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&open=AaDBovNEz6wgCEivzDjz).

It had a loop inside a loop. I pulled the inner loop into its own function, enqueuePropagations. Same logic, same result, just split in two. Quality Gate passes now.

**What I checked before pushing:**
- make build: clean
- make test: all pass
- Both tests that cover this exact function: pass
- A real live migration run: zero errors, and I watched concurrency adjust itself live (25 → 16 workers)

This should show as mergeable with a green Quality Gate now.

mannylee-sonar (approved):
Approving this PR.

**What it does**

SonarQube Cloud stops answering after too many API calls in 5 minutes. This PR adds a rate limiter that holds every request under a set cap, plus a concurrency setting that adjusts itself every 30 seconds based on how fast the API is answering.

**Build and tests**

| Check | Result |
|---|---|
| make build | Clean |
| make test | 17 packages, 0 failures |
| lib/sq-api-go tests (separate module) | All pass |
| make smoke-fast | All pass, including the new --api_max_rate_per_min cases |

**Live test 1: juice-shop (small, 7 languages)**

| Metric | Value |
|---|---|
| Duration | 1m16s |
| Errors | 0 |
| Concurrency change | 25 → 16 workers (avg latency 615ms, target 1500/min) |
| Lines of code, source vs target | 34,666 vs 34,666, exact match |

**Live test 2: stress test, 24,000 issues in one run**

| Metric | Value |
|---|---|
| Duration | 1m07s |
| Errors | 0 |
| Rate-limit rejections (HTTP 429) | 0 |
| Concurrency change | 25 → 17 workers (avg latency 642ms, target 1500/min) |
| Issues migrated | 12,000 + 12,000 = 24,000, exact match both sides |
| Lines of code, both projects | 12,240 and 7,320, exact match both sides |

**One old bug found, not caused by this PR**

Running transfer twice on the same project fails on the second run, with a report-ordering error. I checked: this also fails on unchanged main. Not this PR's fault, already on record from an earlier PR review.

**Fixes I pushed to this branch**

- Merged main in to clear the 4 files GitHub reported as conflicting. Not real conflicts, a rebase artifact. No code actually changed as a result.
- Split one function SonarQube Cloud flagged for being too complex (Cognitive Complexity 18, limit 15) into two smaller ones. Quality Gate is green now.

**Bottom line**

Rate limiter and self-adjusting concurrency both work, proven live, at both a small scale and a 24,000-issue scale, with zero errors and zero rate-limit hits either time. Approved.

PR #587, opened 2026-09-17
Add --insecure flag for self-signed SonarQube Server certificates

### LINKED ISSUE #586: Allow Insecure HTTPS for locally hosted SonarQube Servers
This comes from a customer - if we could allow an `--insecure` flag to allow it to call their local SonarQube which does not have a valid SSL cert. 

This is up for discussion and consideration.

++ @okorach-sonar

joshua-quek-sonarsource:
Closes #586

### What's the problem?

A customer runs SonarQube Server on their own network with a "self-signed" security certificate — think of it as an ID card they printed themselves instead of one issued by an official authority. Our tool refuses to talk to servers like that, so they simply couldn't migrate.

### What does this change?

A new `--insecure` option that tells the tool "I know this server, don't check its ID card." That's it.

### Is this safe?

Three deliberate guardrails:

1. **Off by default.** If you don't type `--insecure`, nothing changes at all — the tool still checks certificates exactly as before.
2. **Only works one direction.** It applies solely to reading *from* the customer's own server. It cannot be turned on for SonarQube Cloud, because Cloud always has a proper certificate — so skipping the check there could only ever hide a real attack.
3. **It tells you.** Every run using it prints a warning explaining the risk, so nobody forgets it's on.

### How do we know it works?

Tests spin up a fake server with a bad certificate and confirm both halves: the tool **accepts** it with `--insecure`, and still **rejects** it without. That second test is the important one — it means this option can never quietly become the default and switch off security for everybody.

Full smoke suite passes, plus the new checks were added to it as required.


joshua-quek-sonarsource:
**Review feedback fixed (plain English):**

The bot spotted two gaps, both now closed:

1. **A real bug.** The tool has a point-and-click wizard (used by `wizard` and `gui`). It reads from the customer's server too — but it was ignoring the new setting, so someone using the wizard would still be blocked even after switching the option on. Worse, when it hit the certificate error it only offered to ask for a *different* certificate, which doesn't help this case at all. The setting now carries through to the wizard properly, with the same warning shown on screen.
2. **Missing docs.** The option was only written up on the main config page. Anyone reading the `transfer` or `sync-issues` page couldn't find it. Added there too.

Still safe by default: a new test confirms that if nobody turns the option on, the wizard keeps checking certificates as before.

Re-ran everything — all tests and the full smoke suite pass. Earlier checks were already green (SonarQube quality gate passed with 0 new issues, 86% coverage on new code).

mannylee-sonar (commented):
Human comment here: Reviewing only - not approving!

**What this PR fixes**

The tool has a new `--insecure` flag (`source.insecure` in config). It skips the TLS certificate check when talking to the source SonarQube Server. This is for a trusted internal server with a self-signed certificate. It is off by default. It only applies to the source, never the target. The tool warns every time it is turned on. This also fixes a bug (#586) where the wizard and the GUI silently dropped this setting instead of using it.

**What I tested**

I built the tool from this PR's branch. I ran the full unit test suite (`make test`), plus two extra test areas the Makefile does not run: the `lib/sq-api-go` module tests, and the `-tags smoke` TLS test. All passed.

I ran a real, live transfer of the `juice-shop` project from a real SonarQube Server to a real SonarQube Cloud sandbox org, using this PR's build. I checked the result on the target directly, not just the tool's own output.

| Claimed fix | Verified | Evidence |
|---|---|---|
| `--insecure` skips the cert check | Yes | Real TLS handshake test against a self-signed cert, passes |
| Off by default, source-only | Yes | Default is `false`; no `target.insecure` field exists |
| Warns every time it is on | Yes | Warning text checked in test |
| Wizard/GUI now passes the setting through (fixes #586) | Yes | Direct unit tests for this path pass |
| Docs and schema updated | Yes | Checked all 3 doc files and the schema by hand |
| CLI flag overrides config file both ways | Yes | Test covers "config true, CLI false" case |

**Problems found**

Running the same transfer twice in a row fails on the second run, with a report-processing error about commit order. I checked: this PR does not touch that code. This looks like a bug that already exists on `main`. I recommend a separate ticket for it. It does not affect this PR's TLS fix.

**Recommendation: Approve.**

Every claimed fix has a passing test or a direct code check. The one problem found is unrelated to this PR's changed files.

joshua-quek-sonarsource:
**Round 2 of review feedback fixed (plain English):**

The bot caught something genuinely embarrassing: our troubleshooting guide was giving the *wrong* advice for this exact problem. If you hit a certificate error, that page told you to use the "mTLS" options — but those prove *your* identity to the server, they don't make the tool trust the server's certificate. So the one page you'd land on after this failure sent you down a dead end.

Rewritten to separate the two situations clearly, and it now points at the new option. Also added the setting to our example config file so it's discoverable.

**All checks green** on the latest code: tests pass, and the SonarQube quality gate passed with 0 new issues.

One housekeeping note: `main` had moved on since this branch started, which put the PR in a conflicted state — and GitHub silently refuses to run CI on a conflicted PR, which is why checks looked stuck earlier. Rebased onto latest `main`, conflicts resolved, everything re-verified and re-run (full smoke suite included).

joshua-quek-sonarsource:
Thanks @mannylee-sonar for helping to review this.
Branch has been rebased, and ready to merge.

joshua-quek-sonarsource:
### Summary for leadership

**The problem:** a customer hosts SonarQube Server on their own network with a self-signed security certificate — like an ID card they printed themselves. Our tool refused to trust it, so they couldn't migrate at all. No workaround existed.

**The fix:** a new `--insecure` option meaning "I know this server, skip the ID check." Available on all four paths that read from a customer's server (`extract`, `transfer`, `sync-issues`, and the point-and-click wizard).

**Why it's safe:**
- **Off by default** — if you don't ask for it, nothing changes.
- **Cannot be used for SonarQube Cloud**, which always has a valid certificate. Skipping the check there could only ever hide a real attack, so it's structurally impossible.
- **Warns on every run** that uses it.
- A test proves we still **reject** bad certificates when the option is off — so this can never quietly become the default.

**Status:** all checks green, SonarQube quality gate passed with 0 new issues. Ready for review.

Closes #586

okorach-sonar (approved):
LGTM

Issue #583, opened 2026-09-17
Provide a means to chose branches that will be migrated, based on the branch last analysis date


okorach-sonar:
Migrating all project branch is not always a desirable outcome:
- Some projects may have a lot of branches that were never cleaned up, although they are not really needed anymore. The migration window is a good moment to perform this housekeeping on the fly
- The more is migrated, the longer the migration and the more stress on the SonarQube Cloud API this creates

For this reason we want to be able to select branches to be migrated, and exclude branches whose last analysis date is older than an given date. So:
- [ ] The migration tool `extract`, `migrate` and `transfer` command should accept a parameter `--branch_analyzed_after` that will only select branches that were analyzed **on or after** the given date, during `extract`, `transfer` or `migrate`
- [ ] The date should be specified in ISO `YYYY-MM-DD` format. If the format is incorrect, the tool should abort with an explicit error on the date format
- [ ] If `--branch_analyzed_after` is not specified, all branches are selected for `extract`, `transfer` or migrate`
- [ ] If both `--branch_analyzed_after` and `--branches` are specified, only branches that meet BOTH conditions are selected for `extract`, `transfer` or migrate` (see #582)
- [ ] The `branch_analyzed_after` can be specified differently in `extract` and `migrate`. Eg it should be possible to `extract` all branches to have them all, and decide which ones to migrate only during `migrate` phase, or, if certain of the branches, only `extract` the desired branches for extraction speed.
- [ ] `--branch_analyzed_after` can also be specified in the configuration file for instance `"branch_analyzed_after": "2025-01-01"`, either at top level (same date will apply to source extract and target migrate) or in the `source` and `target` config sections separately
- [ ] When the `--branch_analyzed_after` setting selects no branch (because they are all older than the given date), the project main branch (whatever its name) should still be selected for `extract`, `transfer` or `migrate`. The reports should mention that the project main branch was migrated despite not meeting the last analysis criteria 
- [ ] If the `branch_analyzed_after` option specifies a date that is more than 2 years (730 days), a warning should be displayed in the logs that the setting may not filter branches much
- [ ] If the `branch_analyzed_after` option is not specified, all branches are selected for `extract`, `transfer` or `migrate`
- [ ] The `branch_analyzed_after` CLI and config file option should be documented

Examples:
```
# 1.
# Extracts all branches
sonar-migration-tool extract
...
# Migrates only branches analyzed on or after 2025-01-01
[...]

PR #595, opened 2026-09-22
fix(migrate): skip branches the target already holds on a re-run (#588)

### LINKED ISSUE #588: Running `transfer` twice on a project fails.
Running the `transfer` command on a project fails as per:
https://github.com/SonarSource/sonar-migration-tool/pull/587#pullrequestreview-5244305803

Thanks @mannylee-sonar for helping to discover this.

mannylee-sonar:
## What was broken

Running `transfer` (or `migrate`) a second time on the same project failed.

The first run worked. The second run died in the Compute Engine with:

```
Report for commit 'b742fef3…' can't be processed: a newer report has already been
processed, and processing older reports is not supported. The last processed report
was for commit 'b7b85b76…'.
```

When the main branch failed this way, every remaining branch was skipped, so the
re-run migrated nothing and the project came out as Partially Migrated.

The commit ids in that message are a red herring. They are the synthetic
`scmRevisionId` values `BuildMetadata` stamps into every report, so they differ on
every run and carry no ordering information. The date is what the CE compares.

The real cause is #557: every branch's report is backdated to the source branch's
real last-analysis date, not to the migration run's own timestamp. A re-run against
an unchanged source therefore rebuilds a report carrying exactly the same date, and
the CE refuses a report dated at or before the one it already processed. Nothing
guarded against this across runs — `loadCompletedBranches` only resumes *within* a
run, because its JSONL lives in the run directory, which is new each time.

Closes #588

## What changed

Before importing a branch, `importAndRecordBranch` now checks the analysis date
SonarQube Cloud already holds for that branch. If it is at or after the date this
run would submit, the branch is skipped.

The source has not changed, so the report would have been identical. Skipping is
therefore lossless. The alternative, nudging the date forward past the target's
latest, would push a near-identical duplicate analysis onto the branch's Activity
on every re-run and throw away the true source date that #557 deliberately
preserves.

Details worth knowing:

- **No extra API call.** The task already calls `project_branches/list` per project
  to learn the target's main branch name. `fetchSCMainBranch` was split into
  `fetchTargetBranches` + `mainBranchName`, and the same response now also yields
  the per-branch analysis dates.
- **The check runs ahead of history replay (#554).** Every history point is older
  than the current snapshot, so replaying onto an already-populated branch would be
  rejected the same way.
- **A skipped branch is recorded as `up_to_date`.** This is deliberately neither
  `success` nor `skipped`. `collectProjectData` uses worst-outcome-wins, so
  recording `skipped` would report a fully migrated project as Skipped purely
  because the operator ran the command twice. `collectProjectData` now treats
  `up_to_date` as a success.
- **Unchanged paths:** a branch the target has never analyzed, a source branch that
  was never analyzed at all (its report is stamped "now", which always wins), a
  first migration into an empty organization, and a source branch that was
  re-analyzed since the last run — including a *new* branch appearing on the source
  after the project was already migrated once (see multi-branch testing below).

## How it was tested

Unit: 8 new tests. `make build`, `make test` (17 packages) and `make smoke-fast` all
pass.

Live, against a real SonarQube Server source and the `mannylee-migration-unbound`
sandbox org on SonarQube Cloud, in three rounds:

**Round 1 — the fix itself**, `juice-shop`, `--skip_issue_sync`:

| Run | Binary | What | Result |
|---|---|---|---|
| A | `main` | fresh import | success |
| B | `main` | re-run | **FAILED**, the CE error above |
| C | fixed | re-run on B's already-failed target | success, branch `up_to_date` |
| D | fixed | project deleted, fresh import | success, 551 issues on target |
| E | fixed | re-run | success, branch `up_to_date` |

**Round 2 — full issue/hotspot sync**, `juice-shop`, sync ON (no `--skip_issue_sync`):

| Run | What | Result |
|---|---|---|
| 1 | Fresh import + sync | success. 551 issues, 28/28 hotspots tagged `sqs-hotspot` + `metadata-synchronized`. |
| 2 | Re-run + sync | branch `up_to_date`; sync tasks ran again with **no duplicates** — same issue/tag counts, same comment counts, still 1 analysis. |

**Round 3 — multi-branch**, `juice-shop` with 2 new branches added on the source
after Round 2's target already existed:

| Run | What | Result |
|---|---|---|
| 3 | Transfer again | main branch `up_to_date` (skipped); both new branches submitted and succeeded |
| 4 | Re-run | all three branches now `up_to_date` |

**Round 4 — rebase compatibility.** This branch was rebased onto a newer `main`
that landed `#583` (`--branch_analyzed_after`) in between, which touches the same
per-project import path (it extracted the goroutine body into `importProjectDataOne`
and added its own branch filter next to `fetchSCMainBranch`). The two features are
independent — `#583` decides which branches are attempted, this fix decides whether
an attempted branch needs resubmitting — so the rebase was a textual conflict, not a
behavioral one. Re-ran `make build`/`make test`/`make smoke-fast` after resolving it
(all pass), then re-ran `transfer` against the target from Round 3 twice more: the
first run unexpectedly resubmitted every branch, traced to the sandbox project having
been reset by something outside this session (a new internal component ID, so
`createProjects` legitimately started fresh); the second run, against the now-stable
project, correctly reported all three branches `up_to_date` again.

Checked on the target after every round, not just the tool's own output:
`api/project_analyses/search` (exactly 1 analysis per branch, always), issue/tag
counts via `api/issues/search`, and `api/project_branches/list`.

### Before and after (Round 1)

| | Run B (before) | Run E (after) |
|---|---|---|
| Branch record | `status=failed` | `status=up_to_date` |
| `importProjectData` | `succeeded=0 failed=1` | `succeeded=1 failed=0` |
| Duration | 25s | 0.6s |

## What reviewers should look at closely

- **The skip is date-based, not content-based.** If a source branch is re-analyzed
  but its last-analysis date somehow does not move forward, the re-import is
  skipped. I judged that acceptable because the CE would reject the submission
  anyway.
- **`up_to_date` in `collectProjectData`.** It is added to the same `case` as
  `success`. Worth confirming that is where you want it.
- **A partially broken first run will not self-heal.** If run 1 imported a branch
  badly but the CE accepted it, run 2 now skips instead of retrying. The escape
  hatch is to delete the target project or branch and run again. There is no
  `--force` flag, by choice, to keep this change small.
- **Overlap with `#583`'s branch filtering.** `importProjectDataOne` (from `#583`)
  and this fix's `fetchTargetBranches`/`targetBranchUpToDate` sit right next to each
  other in the same file. They don't interact — one picks which branches to attempt,
  the other decides whether an attempted branch needs resubmitting — but worth a
  second look given how close they are textually.

## Out of scope

- The synthetic random `scmRevisionId` itself.
- A `--force` re-import flag.
- The generic `There was an issue whilst processing the report` CE error.
- **A re-run still reports "Partial Migration"**, but for an unrelated, pre-existing
  reason: `createProjects` returns 400 `key already exists`, classified "Already
  present" / by-design. It behaves identically on `main` and does not affect the
  exit code.
- Deep issue-sync triage (won't-fix + comment round trip): the test project used
  has no manually-triaged issues to exercise that path with. Sync itself was
  proven to run cleanly and stay idempotent alongside this fix.

## Smoke suite

No smoke change. This adds no command, flag, config field, migration phase, task
group, or written artifact — only a new status value inside the existing
`importProjectData` JSONL. A tier-2 "run the pipeline twice" case would be a
reasonable follow-up guard.


okorach-sonar (approved):
LGTM

